PR #2, opened 2018-02-26
Adds RBAC config for Helm

kerin:
Helm requires a `ClusterRoleBinding` in RBAC-enabled cluster, included in this PR.


SteveMarshall:
To clarify: this PR makes Helm's server component (`tiller`) a cluster admin, yeah?

kerin:
@SteveMarshall yes - that's not something we'd want to do long-term, but it lets us make use of Helm right now

kerin:
[more info](https://github.com/kubernetes/helm/blob/master/docs/securing_installation.md)

PR #3, opened 2018-02-26
Configure nginx-ingress SSL handling and DNS

kerin:
We want to use nginx-ingress to handle all inbound traffic, with SSL termination at its ELB, forcing HTTP->HTTPS redirection, preserving source IP address, and with a wildcard DNS entry (*.apps.cluster) pointing to the ELB. This PR:
    
- adds an ACM certificate with Terraform for attachment to the ELB, using DNS request validation 
- adds a HTTP->HTTPS redirect to all applications' nginx server{} directive
- adds a DNS record annotation for the Service ELB, for use by external-dns
- adds an annotation containing an ACM ARN (cert must already exist and match the DNS name specified for external-dns)
- configure the ELB to handle SSL traffic on port 443 only
- configure the ELB to use non-SSL HTTP for instance traffic (so SSL is terminated on the ELB)
- configure the ELB to also send non-HTTP traffic to nginx-ingress, so it can be redirected to HTTPS/443
- configure the ELB to preserve the clients' source IP address
- creates a service account for nginx-ingress, as the cluster has RBAC enabled

### DIFF
--- a/cluster-components/nginx-ingress/helm-values.yml
+++ b/cluster-components/nginx-ingress/helm-values.yml
@@ -1,0 +1,39 @@
+controller:
+  replicaCount: 3
+
+  # Add a snippet to each nginx server{} block, forcing HTTP->HTTPS
+  # redirection regardless of Ingress rule config
+  config:
+    server-snippet: |
+      if ($http_x_forwarded_proto != 'https') {
+        return 308 https://$host$request_uri;
+      }
+
+  # nginx-ingress's ServiceType=LoadBalancer by default, so we're mainly
+  # configuring an ELB here
+  service:
+    annotations:
+      # Have `external-dns` create a Route53 record pointing to this ELB
+      external-dns.alpha.kubernetes.io/hostname: "*.apps.cloud-platforms-sandbox.k8s.integration.dsd.io"
+
+      # Attach pre-existing ACM cert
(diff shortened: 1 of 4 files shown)


cluster-components/nginx-ingress/helm-values.yml
```diff
@@ -0,0 +1,39 @@
...
+
+  # nginx-ingress's ServiceType=LoadBalancer by default, so we're mainly
+  # configuring an ELB here
+  service:
+    annotations:
+      # Have `external-dns` create a Route53 record pointing to this ELB
+      external-dns.alpha.kubernetes.io/hostname: "*.apps.cloud-platforms-sandbox.k8s.integration.dsd.io"
+
+      # Attach pre-existing ACM cert
+      service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:eu-west-1:926803513772:certificate/d737162e-7dc3-434b-ba0c-b0d9fcb53e7c"
```
SteveMarshall: Do we care if this specific cert goes away? Is this the cert that's created in c949106? If so, can this get the ARN automatically somehow (assuming that's not super overkill…)?

kerin: Yes we care - without the SSL cert we can't do SSL.

Yes it's the cert created in this PR.

No we can't create it automatically - neither Kubernetes nor nginx-ingress have any mechanism for ACM provisioning or attachment other than passing the full ARN of a pre-existing ELB.

PR #6, opened 2018-02-28
Make GitHub webops team members cluster admins

kerin:
Maps github WebOps group members to cluster admin role


SteveMarshall:
@kerin This might be clearer if the commit subject was something like `Make GitHub webops team members cluster admins`? (And probably it doesn't need a commit body then, too.)

PR #8, opened 2018-02-28
Configure Danger to review pull requests

SteveMarshall:
We should use [our existing Danger configuration](https://github.com/ministryofjustice/danger) to make sure our pull requests and commits are in line with the MOJ's standards (as much as we can, anyway).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+script:
+  - gem install ministryofjustice-danger --version '~> 0.1' && danger
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,0 +1,1 @@
+danger.import_dangerfile(github: 'ministryofjustice/danger')


.travis.yml
```diff
@@ -0,0 +1,2 @@
+script:
+  - rbenv global 2.3.1 && gem install ministryofjustice-danger --version '~> 0.1' && danger
```
kerin: What does this actually do? At the moment all PRs result in a failed Travis build because there's no Rakefile. Should there be a Rakefile? What would be in it?

SteveMarshall:
@kerin There shouldn't be a Rakefile. That's what Travis defaults to looking for if it can't find a `.travis.yml`. They're all failing because this hasn't been merged yet but I enabled it so that people could see what this PR does. [Our Danger config](https://github.com/ministryofjustice/danger) is set to [lint commit messages](https://github.com/jonallured/danger-commit_lint#usage) and not much else right now.

PR #11, opened 2018-03-19
First attempt at deploying Concourse in Kube

razvan-moj:


lukaszraczylo:
@razvan-moj ^ could you please correct the following?

PR #22, opened 2018-04-23
Feature/cluster creation pipeline

ale-novo:
cluster creation pipeline changes


bin/create_cluster.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+if [ $# -eq 1 ];
+then
+  workspace="$1"
+  terraform init
```
kerin: I don't think this should be interacting with Terraform directly, as AWS infrastructure-level changes should be handled in a separate phase that will run prior to creating or updating clusters. At a high level, we should have the following phases when creating or updating environments:

1. Terraform - update or create main AWS resources such as VPC, subnets, Route53, etc
2. Kops - update or create cluster
3. Helm/kubectl - update or create services on the cluster

The work here should be solely concerned with item no. 2 above, and at a high level do something like this:

1. `terraform output -json` - get values from the current Terraform state that Kops needs to know about
2. `create_cluster_config.py` - pass those values into `create_cluster_config.py` on the command line, (either as individual flags like `--availability-zones`, or the whole tf json object for simplicity's sake), combine it with a template `cluster-template.yml`, and output a final `cluster.yml`
3. `kops create|replace -f cluster.yml` - apply the changes to kops remote config store
4. `kops update` - apply kops remote config to the actual cluster

We don't necessarily need to wrap numbers 1, 3 and 4 above in scripts, unless it's easier to do so, as they could just be steps in our pipeline.

bin/create_cluster.sh
```diff
@@ -0,0 +1,26 @@
...
+  terraform apply ~/plan.out
+
+  KOPS_STATE_STORE=`terraform output kops_state_store`                                                                                                         
+  CLUSTER_DOMAIN=`terraform output cluster_domain_name`
+  mkdir ~/key/ && chmod 700 ~/key/
+  echo -e  'y\n' | ssh-keygen -t rsa -f ~/key/"$CLUSTER_DOMAIN"_kops_id_rsa -N '' -C "$CLUSTER_DOMAIN" && chmod 600 ~/key/"$CLUSTER_DOMAIN"_kops_id_rsa
+  python ../../bin/create_cluster_config.py ../../kops/"$workspace".yaml > ~/clusterspec.yaml
+  cat ~/clusterspec.yaml
+  kops create -f ~/clusterspec.yaml --state="s3://$KOPS_STATE_STORE" || exit 1
+  kops create secret --name "$CLUSTER_DOMAIN" --state="s3://$KOPS_STATE_STORE" sshpublickey admin -i ~/key/"$CLUSTER_DOMAIN"_kops_id_rsa.pub
```
kerin: I think it might be possible to also specify the SSH key in kops YAML, but I need to check the docs on that

bin/create_cluster_config.py
```diff
@@ -0,0 +1,183 @@
...
+    print('error executing terraform command, exit.')
+
+# Populate variables from terraform output
+cluster_name = tf('cluster_domain_name')
+dns_zone = cluster_name
+kops_state_store = 's3://' + tf('kops_state_store') + '/' + cluster_name
+availability_zones = tf('availability_zones')
+master_public_name = 'api.' + cluster_name
+network_cidr = tf('network_cidr_block')
+topology = 'bastion.' + cluster_name
```
kerin: It looks like this is actually the bastion name, not a value for `topology` (which on the kops command line can only be `public` or `private`) - although we likely don't need to specify bastions here at all (see below)

bin/create_cluster_config.py
```diff
@@ -0,0 +1,183 @@
...
+    else:
+        sys.exit(key + ' not found, exit.')
+
+
+# Dumper config
+yaml.SafeDumper.add_representer(literal, literal_presenter)
+yaml.SafeDumper.ignore_aliases = lambda *args: True
+
+# Parser config
+p = pprint.PrettyPrinter(indent=4)
```
kerin: This doesn't seem to be used anywhere

bin/create_cluster_config.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/python
```
kerin: We should use Python 3, and it looks like this script is, so the shebang should be `#!/usr/bin/env python3`

bin/create_cluster_config.py
```diff
@@ -0,0 +1,183 @@
...
+internal_subnets = tf('internal_subnets')
+external_subnets = tf('external_subnets')
+
+# Organize in lists by kind and role
+for template in templates:
+    if template['kind'] == 'Cluster':
+        clusters.append(template)
+    if template['kind'] == 'InstanceGroup':
+        instances.append(template)
+        if template['spec']['role'] == 'Bastion':
```
kerin: It's probably better to leave bastions out of this, and create those in Terraform, as bastion hosts aren't really related to the cluster, and we might want more control over them than Kops allow.

bin/create_cluster_config.py
```diff
@@ -0,0 +1,183 @@
...
+            bastions.append(template)
+        if template['spec']['role'] == 'Node':
+            nodes.append(template)
+        if template['spec']['role'] == 'Master':
+            masters.append(template)
+
+# Update all Cluster kind
+for template in clusters:
+    template['metadata'].update({'name': cluster_name})
```
kerin: General python best practice note - it's not necessary to use the `update` method when setting a single key in a dict - `template['metadata']['name'] = cluster_name` does the same thing, and is shorter and easier to understand. `update()` is best used for setting several dict keys at once, so you should either use one `update()` call with several values, or directly set keys individually

kops/cloud-platforms-test.yaml
```diff
@@ -0,0 +1,120 @@
...
+  additionalPolicies:
+    node: |
+      [
+        {
+          "Effect": "Allow",
+          "Action": [
+            "route53:ChangeResourceRecordSets"
+          ],
+          "Resource": [
+            "arn:aws:route53:::hostedzone/Z1N3S2UZYKP74S"
```
kerin: The Route53 hosted zone ID is a per-environment value, so this should also be populated by the `create_cluster_config` script

kops/cloud-platforms-test.yaml
```diff
@@ -0,0 +1,120 @@
...
+      ]
+  authorization:
+    rbac: {}
+  channel: stable
+  cloudProvider: aws
+  iam:
+    allowContainerRegistry: true
+    legacy: false
+  kubeAPIServer:
+    oidcClientID: 7ix4YkYwQe8n7j9qTrdAKeMz26eJTB2x
```
kerin: `oidcClientId` and `oidcIssuerURL` will also vary per cluster, so should be set by the script

kops/cloud-platforms-test.yaml
```diff
@@ -0,0 +1,120 @@
...
+---
+
+apiVersion: kops/v1alpha2
+kind: InstanceGroup
+metadata:
+  creationTimestamp: null
+  name: master-eu-west-1a
+spec:
+  image: kope.io/k8s-1.8-debian-jessie-amd64-hvm-ebs-2018-01-14
+  machineType: t2.medium
```
kerin: We should probably also have instance types and numbers as arguments in the cluster creation script, so we can deploy smaller clusters for testing and individual dev clusters

kerin:
Also added @kalbir as a reviewer, given the scope of what's happening here and the implicit design decisions within - and maybe others should take a look too?

bin/create_clusters.py
```diff
@@ -0,0 +1,64 @@
...
+
+# create the corresponding clusters
+for item in yaml:
+    if item in workspaces and item in clusterlist:
+        print('cluster ' + item + ' already exists, skipping')
+    else:
+        print('creating cluster ' + item)
+        run('../../bin/create_cluster.sh ' + item)
+
+print('Done.')
```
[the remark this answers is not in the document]
kerin: After a very brief chat with @kalbir just now I think this probably warrants some discussion - in terms of how we handle the config of multiple clusters, how/when those configs are applied, and what testing or previewing config changes looks like. We could have individual `cluster.yml` files for each individual cluster, but then that raises the question of what should be in there and what should be passed in at YML render time - ideally we want to be able to test cluster config changes without needing to apply them to a specific running cluster, so we'd need to be careful to keep things like cluster and DNS names out of it.

The other question is if we should have a single script or process that applies all changes to all clusters in one go - that doesn't feel right to me as it'd require a one-to-one mapping between `cluster.yml` files and deployed clusters, preventing us from previewing or testing changes, but I'm open to suggestions here.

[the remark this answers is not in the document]
ale-novo: Kerin, if you have a pipeline that creates/updates a specific cluster, you cannot run that for multiple clusters, because its designed to work with a specific cluster.
You can have however a generic pipeline that takes as an argument the cluster name and creates/updates that given cluster, and then you can reuse that pipeline for different clusters.
we dont want to reuse the same pipeline code over and over because that will generate code duplication.

the one to one mapping between cluster.yaml and the deployed clusters is what we already have on this repo (sandbox cluster and non-prod cluster), inside kops, so that is why is used that approach for this feature.

again, we can have a generic pipeline manually triggered from witch you specify as argument the cluster name, but the issue with that is that you cant guarantee the cluster.yaml in the repo will match the running cluster config. The running cluster config may be outdated if the pipeline did not run for that given cluster after a commit was pushed to master.
with this approach everything that is in the repo is applied on the cluster and you can consider the repo the source of truth.
You can have a 'testing' branch of this repo to apply it against a testing subset of clusters to test the config before applying it to the production clusters.

bin/create_cluster.sh
ale-novo: Hi, this work in jira was divided into 2 tkts, terraform elements and kops elements
https://dsdmoj.atlassian.net/browse/CPT-472
https://dsdmoj.atlassian.net/browse/CPT-483

regarding the terraform part specifically:

1)Use terraform to provision the VPC and other infrastructure needed.
2)Turn this into an automated job to kick off the terraform build
3)Make it so that any variables required by the kops job in CPT-483 can be passed to it.

that is what create_cluster.sh is doing.
If this is not what we want, please update the described tkts with the correct specification, and ill work on modifying those.

Your description above is just a simplified version of what this pull request is already doing.
If we already have and automated way of creating a cluster, all the involved components, terraform workspace, terraform components (vpc and networking), and kops elements automatically , why turn that into manual steps that require us to handle them?

bin/create_cluster.sh
ale-novo: ok

bin/create_cluster_config.py
ale-novo: yes good catch i missed that.

bin/create_cluster_config.py
ale-novo: yes, agree

bin/create_cluster_config.py
ale-novo: This was the way we did it for sandbox and nonprod clusters that already exist. Bastions are created by kops.
When you say to create those in terraform, im assuming using kops to generate the terraform template for the bastions and then executing that template?

bin/create_cluster_config.py
ale-novo: sounds good

bin/create_cluster.sh
kerin: The tickets are correct as far as I can see, and they don't describe handling Terraform as part of cluster creation:

- Use terraform to provision the VPC and other infrastructure needed.
- Turn this into an automated job to kick off the terraform build
- Make it so that any variables required by the kops job in   CPT-483 READY FOR REVIEW  can be passed to it.

Terraform handles management of all AWS resources that _aren't_ the cluster, so there's a separation between what Terraform and Kops do - therefore it makes sense to handle them as separate phases with separate processes and scripts managing them.

kops/cloud-platforms-test.yaml
ale-novo: this is a tricky one, and decided to hardcode it to get this working for the sprint.
the problem is that you have to run the terraform outputs for the terraform/global-resources to populate that value, and the terraform variables files is encrypted, making it not work form inside the code-build environment. 
We need to think a way of exporting the key so its possible to de encrypt the repo from within the pipeline. Totally out of scope for this sprint.

kops/cloud-platforms-test.yaml
ale-novo: by create_cluster_config.py?

kops/cloud-platforms-test.yaml
ale-novo: or we can have a testing branch of this repo with the smaller instance size, and link that branch to a create-cluster-test-pipeline job that will run that against test cluster for test purposes before merging to master (prod)

kops/cloud-platforms-test.yaml
kerin: Yep - although, as per the discussion on how we should manage per-cluster config, maybe this actually should be a static value in an individual cluster's YAML file, so I'd leave this for now

bin/create_cluster.sh
ale-novo: If that is the case, we can remove that from the create_cluster.sh
This PR also includes changes to the terraform templates to create the VPC and underlying networking parts, from terraform workspaces. So if you create a new terraform workspace and apply the templates it will create the vpc and other parts for the new cluster.
This will be an extra manual step if we remove it from create_cluster.sh

kops/cloud-platforms-test.yaml
ale-novo: ok ill take a look

kops/cloud-platforms-test.yaml
ale-novo: ok

bin/create_cluster_config.py
ale-novo: this was fixed

bin/create_cluster_config.py
ale-novo: this was fixed

bin/create_cluster_config.py
ale-novo: i changed this only for new keys
in cases where the key exists , using that method will override other values inside the key making it inviable
the .update method must be used in those cases

kops/cloud-platforms-test.yaml
ale-novo: if we put instance type and numbers as arguments for the cluster creation script, its not going to live in code, nor in git.
If we put it in the pipeline, then we will need one pipeline per cluster creating duplication.
I will suggest leaving this for now as it is.
the cluster yaml specification fill will have the size and number of instances for each cluster.

kops/cloud-platforms-test.yaml
ale-novo: as we determined to have a per cluster yaml file, this will be static for now.

kops/cloud-platforms-test.yaml
kerin: Yep, agreed - we can pull it out later, or pull the `InstanceGroup` specs into separate YML files later if it proves to be a problem.

bin/create_cluster.sh
ale-novo: this was fixed in https://github.com/ministryofjustice/kubernetes-investigations/pull/22/commits/e81efc201cd5f71f549fffb864a4505d7445b7ab

kerin: What's the context in which `00-cluster_pipeline.sh` is intended to be executed? I don't think we ever want to execute a single command and create both Terraform and Kops resources all in one go, as they're distinctly different things, and will quite likely have significantly different lifecycles - thoughts @kalbir ?

ale-novo: Kerin, 00-cluster_pipeline.sh is just a way of executing the different stepts automatically in a pipeline for 1 cluster.
you can manually execute 01. 02. 03. 04 and accomplish the same

99-multiple_clusters.py is the first script executed by the pipeline, this script determines what clusters need to be created, and then triggers 00-cluster_pipeline.sh for each cluster, witch internally executes the steps for each cluster, 
01-terraform_resources.sh
02-rsa_key.sh
03-cluster_config.py
04-kops_elements.sh

kerin: OK - but that means that the Terraform phase and cluster phase are still bound together, and any change will trigger both `terraform apply` and `kops create`. We don't want that, as they have separate lifecycles - we're likely to have frequent changes to Terraform resources that don't impact the cluster at all, so they shouldn't trigger anything with Kops.

Similarly, the Kops `03-cluster_config.py` script is still interacting with Terraform directly, which binds the two together. Kops does not need to know anything about Terraform, so it would be cleaner to pass information obtained from Terraform into this script as command line arguments, as per [previous feedback](03-cluster_config.py) - this would make the script more flexible, as it wouldn't depend on any extra tooling or particular execution context.

PR #28, opened 2018-05-24
Ecr terraform module documentation

ale-novo:
This documentation is part of bullet point:
`These instructions have been read and tested for clarity and correctness by a third party (not the developer or reviewer).`

of issue
https://github.com/ministryofjustice/cloud-platform-reference-app/issues/29

This document should cover all terraform code within the terraform directory of this repo, however to keep the focus of the issue https://github.com/ministryofjustice/cloud-platform-reference-app/issues/29 it is only centered in the ECR credentials part of it, outlining only the structure for future documentation of other parts.

this document covers instructions for cloud platforms team to create ecr credentials for other teams when requested

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -51,26 +51,9 @@ $ Terraform apply
 ```
 In the same directory where the newly created Terraform file lives.
 
-### Module Inputs
-
-| Name | Description | Type | Default | Required |
-|------|-------------|:----:|:-----:|:-----:|
-| repository_name | name of the repository to be created | string | - | yes |
-| team_name | name of the team creating the credentials | string | - | yes |
-
 **Note that this example may create resources which can cost money. Run `Terraform destroy` when you don't need these resources.**
 
-### Module Outputs
-
-| Name | Description |
-|------|-------------|
-| policy_arn | ARN for the new policy |
-| access_key_id | Access key id for the new user |
-| secret_access_key | Secret for the new user |
-| user_name | User name for the new credentials |
-| repository_arn | ECR repository ARN |
-
-The user_name ,secret_access_key and repository_arn will need to be distributed to the Team after this code is run.
+The user_name ,secret_access_key and repository_arn outputs will need to be distributed to the Team after this code is run.
 
 ## Modules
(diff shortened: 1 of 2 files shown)


terraform/README.md
```diff
@@ -0,0 +1,72 @@
...
+## Cloud Platform
+The cloud-platform directory is were the cluster specific code lives and relies on Terraform workspaces as the starting point.
+
+## Global Resources
+The global-resources directory contains Terraform code to build the common resources for all clusters.
+Within Global Resources you can find:
+
+### ecr_credentials.tf
+
+This Terraform file uses the following Terraform module which creates ECR credentials and repository on AWS. This Module is used to create the user, ECR repository and credentials for our example app. The team name is `webops` and the repository name is `cloud-platform-reference-app` and these are the parameters used by the module in this case
```
alkar: Perhaps it's better to just refer to https://github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials/blob/master/README.md ? Correct me if I'm wrong but this section seems to be duplicating the docs from the repo. I think it would be better to just explain how the user should handle code changes here (make a new file / expand `ecr_credentials.tf`) and leave the details in the ecr module repo.

ale-novo: this is a specific usage of the module in this case for the demo app, i think that should be documented here.

AntonyBishop:
The content reads ok. I'm just not sure that there is a clear "why" or "conditions when" a user would need to do this? This may be a context thing and be related to other guidance but from here I can't tell.

terraform/README.md
```diff
@@ -0,0 +1,72 @@
...
+In the same directory where the newly created Terraform file lives.
+
+### Module Inputs
+
+| Name | Description | Type | Default | Required |
+|------|-------------|:----:|:-----:|:-----:|
+| repository_name | name of the repository to be created | string | - | yes |
+| team_name | name of the team creating the credentials | string | - | yes |
+
+Note that this example may create resources which can cost money. Run `Terraform destroy` when you don't need these resources.
```
AntonyBishop: Might be worth making this bold or stand out otherwise folks might miss it.

ale-novo: this was fixed

ale-novo:
a context was added why this is nedded

PR #30, opened 2018-05-29
create concourse tf state bucket; enable SSE on all state buckets

alkar:


alkar:
This is related to work done in https://github.com/ministryofjustice/cloud-platform/issues/71

PR #35, opened 2018-06-04
component configuration for cloud-platform-live-0

alkar:
Connects to https://github.com/ministryofjustice/kubernetes-investigations/pull/31


cluster-components/cloud-platform-live-0-kuberos/kuberos.yml
```diff
@@ -0,0 +1,93 @@
...
+        server: https://api.cloud-platform-live-0.k8s.integration.dsd.io
+      name: cloud-platform-live-0.k8s.integration.dsd.io
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: kuberos
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
```
kerin: Question that's occurred to me before - does it make sense to declare a specific ingress class when we've only got one ingress controller? If we change ingress controller that means we'll need to update any ingress rules that have declared a class, so is it better to not define anything and use the default controller?

alkar: It very much depends on how we want to handle a transition in the future but I'm not sure there's a clear answer. I would rather not leave it unset since the default behaviour for ingress controller is not necessarily catch-all. At least by having a value, we can better control the transition by adjusting selector flags on the controller. On the other hand, I feel `nginx` is too specific.

kerin: Fair enough - it's not an issue right now, so we can address it later if we switch ingress controller 👍

alkar:
@kerin thanks, that's going to be needed in the future

PR #38, opened 2018-06-07
Change oidcClientID for my test cluster

tatyree:
As per Mitch's notes:
https://gist.github.com/alkar/57d58394dd2d7b773eebbe911472bf2c

I **think** this is the only value that needs changing...


alkar:
@tatyree this needs to go on a separate config file, named after the new cluster - instructions might be a bit unclear!

tatyree:
@alkar fixed.  Can you check it again for me, please?

PR #39, opened 2018-06-07
Delete sandbox cluster

ale-novo:
this pr is to delete any reference to the sandbox cluster, witch was deleted.

issue: https://github.com/ministryofjustice/cloud-platform/issues/113


cluster-components/kuberos/ingress.yml
```diff
@@ -1,15 +0,0 @@
-apiVersion: extensions/v1beta1
```
alkar: It should be safe to remove all the files under `cluster-components/kuberos/`

concourse/Concoursefile.yaml
```diff
@@ -1,51 +0,0 @@
-resource_types:
```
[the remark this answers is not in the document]
ale-novo: github_concourse.ini points to the sandbox cluster, and according to raz this was an investigation and its not being used

cluster-components/kuberos/ingress.yml
ale-novo: ok

ale-novo: fixed

PR #43, opened 2018-06-22
DNS record for ECR registry

alkar:


alkar:
AWS does not support custom domain names for ECR

PR #47, opened 2018-07-10
auth0-kuberos

razvan-moj:
The scope of this PR is to document the steps needed to get to a working OIDC authenticated app in a k8s cluster; once done the README will fulfil #216 (Auth0 used to verify Github identity/membership) and #221 (Kuberos is the app that will use this authentication method)


razvan-moj:
this issue links to #216

razvan-moj:
Connects to #216

auth0/whitelist-github-teams.js
```diff
@@ -0,0 +1,32 @@
+function (user, context, callback) {
+  var request = require('request');
+  var github_team_whitelist = ['webops'];
```
tatyree: This works for our current requirements, but we'll need to refine it soon so that we give a lower level of access to regular users.

PR #76, opened 2018-09-20
Increase live-0 node count

alkar:
Due to the nature of EBS volumes being tied to a single AZ in AWS, when performing a rolling-update in nodes there is a period between terminating a node and when the replacement is launched during which, pods using EBS volumes are unable to schedule.

This change doubles the number of nodes so that there is always at least one available per zone during a rolling update.


alkar:
Changed to maintain the same node size, resizing to be done in https://github.com/ministryofjustice/cloud-platform/issues/408

PR #82, opened 2018-10-30
Refactor the kops manifest creation

jasonBirchall:
connects to https://github.com/ministryofjustice/cloud-platform/issues/458

**WHAT**
1. `kops.yaml.tpl` to template the creation of a kops manifest. When created it will populate with the hosted zone and cluster name. I used the original live0 kops manifest as an example and added additional admission plugins:
```
  enableAdmissionPlugins:
    - Initializers
    - NamespaceLifecycle
    - LimitRanger
    - ServiceAccount
    - PersistentVolumeLabel
    - DefaultStorageClass
    - DefaultTolerationSeconds
    - MutatingAdmissionWebhook
    - ValidatingAdmissionWebhook
    - NodeRestriction
    - ResourceQuota
    - PodSecurityPolicy
```
2. `kops.tf` to create a rendered version of the kops template above and save it in the `kops` dir.

**WHY**
1. This will allow us to replace the current python implementation that does a `terraform output` lookup to populate variables. 

2. The newly created file will allow us to execute `kops cluster create` using the outputted file. Building our cluster as per specifications.


terraform/cloud-platform/templates/kops.yaml.tpl
```diff
@@ -0,0 +1,289 @@
...
+      ]
+  authorization:
+    rbac: {}
+  channel: stable
+  cloudProvider: aws
+  iam:
+    allowContainerRegistry: true
+    legacy: false
+  kubeAPIServer:
+    oidcClientID: uVp00zncGYUyqr5IMkLH3zNpvLbwzzyX
```
jasonBirchall: When https://github.com/ministryofjustice/cloud-platform/issues/459 is complete we can populate the `oidc` entries.

terraform/cloud-platform/kops.tf
```diff
@@ -0,0 +1,22 @@
...
+    hosted_zone_id        = "${module.cluster_dns.cluster_dns_zone_id}"
+    cluster_dns_zone_id   = "${aws_key_pair.cluster.key_name}"
+    kops_state_store      = "${data.terraform_remote_state.global.kops_state_store}"
+    network_cidr_block    = "${module.cluster_vpc.vpc_cidr_block}"
+    internal_subnets_id_a = "${module.cluster_vpc.private_subnets[0]}"
+    internal_subnets_id_b = "${module.cluster_vpc.private_subnets[1]}"
+    internal_subnets_id_c = "${module.cluster_vpc.private_subnets[2]}"
+    external_subnets_id_a = "${module.cluster_vpc.public_subnets[0]}"
+    external_subnets_id_b = "${module.cluster_vpc.public_subnets[1]}"
+    external_subnets_id_c = "${module.cluster_vpc.public_subnets[2]}"
```
jasonBirchall: I can only pass the slice via the above method, for example, `${external_subnets_id_a[0]}` doesn't work.

PR #85, opened 2018-11-05
Automate/fluentd

ollieanwyll:
WHY
Connects to: ministryofjustice/cloud-platform#469


terraform/cloud-platform-components/fluentd-helm/fluentd-es/values.yaml
```diff
@@ -0,0 +1,7 @@
+namespace: logging
+
+version: v2.2.0
+
+FLUENT_ELASTICSEARCH_HOST: "search-cloud-platform-test-o2m2taivvjpovbcl63mlytnpua.eu-west-1.es.amazonaws.com"
+FLUENT_ELASTICSEARCH_AUDIT_HOST: "search-cloud-platform-audit-effm3qdiau42obkarrpvdxioxm.eu-west-1.es.amazonaws.com"
+FLUENT_KUBERNETES_CLUSTER_NAME: cloud-platform-test-1
```
jasonBirchall: This will need to be interpolated either by the cloud-platform state store or perhaps `${terraform.workspace}`

terraform/cloud-platform-components/fluentd-helm/fluentd.tf
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    set {
+        name = "FLUENT_ELASTICSEARCH_AUDIT_HOST"
+        value = "search-cloud-platform-audit-effm3qdiau42obkarrpvdxioxm.eu-west-1.es.amazonaws.com"
+    }
+
+    set {
+        name = "FLUENT_KUBERNETES_CLUSTER_NAME"
+        value = "cloud-platform-test-1"
```
jasonBirchall: The same as above.

terraform/cloud-platform-components/resources/fluentd-es/templates/00-namespace.yaml
```diff
@@ -0,0 +1,13 @@
+#This file creates the logging namespace for flentd-es. It is named as is to ensure it's the first file deployed.
```
alkar: We do not want to manage the namespace in the chart (I don't think any chart we're using does this).

terraform/cloud-platform-components/resources/fluentd-es/values.yaml
```diff
@@ -0,0 +1,8 @@
+namespace: 
+
+version: 
+
+FLUENT_ELASTICSEARCH_HOST:
```
alkar: As discussed, please use camelcase for the values.

terraform/cloud-platform-components/resources/fluentd-es/templates/04-networkpolicy.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: networking.k8s.io/v1
```
alkar: Please remove `NetworkPolicies` as well, we will be managing these in the environments repo.

terraform/cloud-platform-components/resources/fluentd-es/templates/fluentd-es-config.yaml
```diff
@@ -0,0 +1,338 @@
...
+  name: fluentd-es-config-v0.1.4
+  namespace: {{ .Values.namespace }}
+  labels:
+    addonmanager.kubernetes.io/mode: Reconcile
+data:
+  system.conf: |-
+    <system>
+      root_dir /tmp/fluentd-buffers/
+    </system>
+  containers.input.conf: |-
```
alkar: As discussed, please extract these config files like you've done for `system.input.conf`

terraform/cloud-platform-components/resources/fluentd-es/values.yaml
```diff
@@ -0,0 +1,8 @@
+namespace: 
+
+version:
```
alkar: I believe we should use `image.repository` and `image.tag` as most of the other charts we use (see [here](https://github.com/helm/charts/blob/master/stable/nginx-ingress/values.yaml#L6-L8)), and provide a default value that is the "supported" version.

terraform/cloud-platform-components/resources/fluentd-es/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
```
alkar: `appVersion` should match the default value for `image.tag`

terraform/cloud-platform-components/resources/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -0,0 +1,118 @@
...
+  apiGroup: ""
+roleRef:
+  kind: ClusterRole
+  name: {{ .Release.Name }}
+  apiGroup: ""
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: fluentd-es-{{ .Values.version }}
```
alkar: When you create a new chart using `helm` (`helm create foo`) you'll see that the resources have a number of labels defined:
```
  name: {{ include "foo.fullname" . }}
  labels:
    app: {{ include "foo.name" . }}
    chart: {{ include "foo.chart" . }}
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}
```
I've not found any documentation on how these are used by `helm` but it seems they are used. `nginx-ingress` and `kube-prometheus` define them on all resources. I think we should do the same?

terraform/cloud-platform-components/fluentd.tf
```diff
@@ -0,0 +1,29 @@
+resource "helm_release" "fluentd_es" {
+  name  = "fluentd-es"
+  chart = "resources/fluentd-es"
```
alkar: Can we move the chart under `/helm-charts` please?

terraform/cloud-platform-components/resources/fluentd-es/templates/fluentd-es.yaml
ollieanwyll: Where are you referencing this from?

alkar: Try `helm create newchart` and look at the templates.
Also [example from `nginx-ingress`](https://github.com/helm/charts/blob/master/stable/nginx-ingress/templates/controller-deployment.yaml#L4-L11)

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -85,7 +85,7 @@ spec:
       labels:
         k8s-app: fluentd-es
         kubernetes.io/cluster-service: "true"
-        version: {{ .Values.image.tag}}
+        version: {{ .Values.version }}
```
alkar: Regression?

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -0,0 +1,143 @@
...
+roleRef:
+  kind: ClusterRole
+  name: {{ .Release.Name }}
+  apiGroup: ""
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Values.namespace }}
```
alkar: Again, looking at stable charts and the output of `helm create new-chart`, they don't define `namespace` for  resources. We should do the same and omit this on all resources.

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -0,0 +1,143 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}
```
alkar: From our conversation the other day, if you run `helm create newchart` and look at the generated `Deployment` template (or look at [nginx-ingress](https://github.com/helm/charts/blob/master/stable/nginx-ingress/templates/controller-deployment.yaml#L11)) it uses the fullname template (defined in `_helpers.tpl`). We should use the same and keep it consistent for all resources.

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -0,0 +1,143 @@
...
+    spec:
+      priorityClassName: node-critical
+      serviceAccountName: {{ .Release.Name }}
+      tolerations:
+      # This needs to run on all nodes, including masters.
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      containers:
+      - name: {{ .Release.Name }}
+        image: k8s.gcr.io/fluentd-elasticsearch:{{ .Values.image.tag }}
```
alkar: Remove hardcoded value and set as default in `values.yaml`.

terraform/cloud-platform-components/fluentd.tf
```diff
@@ -0,0 +1,29 @@
...
+  name  = "fluentd-es"
+  chart = "../../helm-charts/fluentd-es"
+
+  set {
+    name  = "namespace"
+    value = "logging"
+  }
+
+  set {
+    name  = "image.tag"
```
alkar: This is not required, we are using the default value (and in case of an upgrade should upgrade the default in the chart).

PR #99, opened 2018-11-15
Automate Prometheus Operator

jasonBirchall:
**OVERVIEW**
---
This PR connects to https://github.com/ministryofjustice/cloud-platform/issues/468 and allows the automation of Prometheus operator with Terraform. When merged, the Terraform will need to be applied to all relevant cluster manually using.
```bash
$ terraform init
$ terraform workspace select <clusterName>
$ terraform apply
```
Eventually, we will place this .tf in a pipeline to run on all clusters upon creation. 

**WHAT**
---
The PR contains the following:
1. `helm-charts/kube-prometheus/custom-alerts.yaml` uses the PrometheusRule CRD to install additional cluster triggered alerts.
2. `helm-charts/kube-prometheus/values.yaml.tpl` is a custom set of values required for the upstream Helm chart. Any key value with an empty string will be interpolated by Terraform, either via an upstream state or the `variables.tf`
3. `terraform/cloud-platform-components/prometheus.tf` contains the Terraform required to do the following:
- Add the coreos Helm chart for Prometheus operator.
- Add a storage class for persistent storage.
- Install the prometheus-opertator using the default `values.yaml` provided by coreos.
- Install kube-prometheus using the values file (number 2) and utilising the `set` command in the Terraform Helm provider.
4. `terraform.tfvars` & `variables.tf` allow the `prometheus.tf` file interpolate the pagerduty and slack config. 

**WHY**
---
1. This will allow the cluster to bootstrap custom alerts (that the team have seen fit).
2. We use our own values file to allow deeper customisation of the kube-prometheus chart. Things like Grafana and Slack hooks are configured here. 
3. Enables the automation of Prometheus operator.  
4. The Slack and PagerDuty webhooks are interpolated from the `variables.tf`

**NOT**
---
What does this PR not do:
_This PR doesn't enable frontends for Promtheus, AlertManager and Grafana. This will need to be performed after the installation._


terraform/cloud-platform-components/prometheus.tf
```diff
@@ -0,0 +1,74 @@
+resource "helm_repository" "coreos" {
+  name = "coreos"
+  url  = "https://s3-eu-west-1.amazonaws.com/coreos-charts/stable/"
+}
+
+resource "kubernetes_storage_class" "prometheus" {
```
alkar: Should we have a retain `reclaimPolicy` ?

terraform/cloud-platform-components/prometheus.tf
```diff
@@ -0,0 +1,74 @@
...
+  chart         = "coreos/kube-prometheus"
+  namespace     = "monitoring-terraform"
+  recreate_pods = "true"
+
+  values = [
+    "${file("../../helm-charts/prometheus-operator/kube-prometheus/values.yaml")}",
+  ]
+
+  set {
+    name  = "grafana.extraVars[0].value"
```
alkar: This feels too brittle (what if we re-order elements?) - also the same for the slack configs below.
Should we instead treat `values.yaml` as a template file and use terraform interpolation instead of overriding using `set {}` ?

terraform/cloud-platform-components/prometheus.tf
```diff
@@ -0,0 +1,74 @@
...
+
+  parameters {
+    type = "gp2"
+  }
+}
+
+resource "helm_release" "prometheus_operator" {
+  name          = "prometheus-operator"
+  chart         = "coreos/prometheus-operator"
+  namespace     = "monitoring-terraform"
```
alkar: Do we want to run this in parallel with the existing one?

terraform/cloud-platform-components/prometheus.tf
```diff
@@ -0,0 +1,74 @@
...
+  depends_on = [
+    "null_resource.deploy",
+    "helm_repository.coreos",
+  ]
+}
+
+resource "helm_release" "kube_prometheus" {
+  name          = "kube-prometheus"
+  chart         = "coreos/kube-prometheus"
+  namespace     = "monitoring-terraform"
```
alkar: Same as above.

terraform/cloud-platform-components/prometheus.tf
```diff
@@ -0,0 +1,74 @@
...
+
+  values = [
+    "${file("../../helm-charts/prometheus-operator/kube-prometheus/values.yaml")}",
+  ]
+
+  set {
+    name  = "grafana.extraVars[0].value"
+    value = "https://grafana.apps.${data.terraform_remote_state.cluster.cluster_domain_name}"
+  }
+
```
alkar: From the README:
```
  --set-string grafana.adminUser=$(head -c 16 /dev/urandom | xxd -p) \
  --set-string grafana.adminPassword=$(head -c 16 /dev/urandom | xxd -p)
```
Although we agreed we won't manage the extra bits yet, we still need to override these values. `hex` from [`random_id`](https://www.terraform.io/docs/providers/random/r/id.html) should be equivalent.

terraform/cloud-platform-components/prometheus.tf
jasonBirchall: Yeah, good catch. Will get a commit in now.

terraform/cloud-platform-components/prometheus.tf
jasonBirchall: Yeah, good catch. Will get a commit in now.

jasonBirchall: ce51bd4

terraform/cloud-platform-components/prometheus.tf
jasonBirchall: ce51bd4

terraform/cloud-platform-components/prometheus.tf
jasonBirchall: Not sure I can set a `reclaimPolicy` in tf provider:
https://www.terraform.io/docs/providers/kubernetes/r/storage_class.html#storage_provisioner

terraform/cloud-platform-components/prometheus.tf
jasonBirchall: I think this is fair. The set command is how Helm recommends doing interpolation but we may as well use Terraform. 

I have made a rather big change in 652b1b3 and 3ee0687.

Helm docs:
https://docs.helm.sh/chart_best_practices/values/#consider-how-users-will-use-your-values

helm-charts/kube-prometheus/values.yaml
```diff
@@ -0,0 +1,4 @@
+############################################
+# NOTE: This file is managed by Terraform  #
+# Don't make any manual modification to it #
+############################################
```
jasonBirchall: Probably worth mentioning that this is a placeholder file for `local_file` to overwrite.

terraform/cloud-platform-components/prometheus.tf
jasonBirchall: Okay. This has been done under b88057b.

PR #100, opened 2018-11-15
Fix HAproxy DNS

razvan-moj:
**Why**
https://cbonte.github.io/haproxy-dconv/1.9/configuration.html#5.3

### DIFF
--- a/haproxy/userdata.sh.tpl
+++ b/haproxy/userdata.sh.tpl
@@ -6,7 +6,6 @@ apt-get -y install haproxy
 tee -a /etc/haproxy/haproxy.cfg <<EOF
 
 resolvers awsdns
-  parse-resolv-conf
   resolve_retries       3
   timeout resolve       1s
   timeout retry         1s
--- a/haproxy/userdata.sh.tpl
+++ b/haproxy/userdata.sh.tpl
@@ -6,7 +6,7 @@ apt-get -y install haproxy
 tee -a /etc/haproxy/haproxy.cfg <<EOF
 
 resolvers awsdns
-  nameserver aws ${var.dns_ip}:53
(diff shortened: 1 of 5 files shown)


haproxy/variables.tf
```diff
@@ -50,3 +50,11 @@ variable "backends_weights" {
     },
   ]
 }
+
+variable "cidr_block" {
+  default = "100.127.0.0/16"
+}
+
+variable "dns_ip" {
+  default = "100.127.0.2"
```
alkar: You can also use this https://www.terraform.io/docs/configuration/interpolation.html#cidrhost-iprange-hostnum-

haproxy/userdata.sh.tpl
```diff
@@ -5,6 +5,18 @@ apt-get -y install haproxy
...
+  nameserver aws ${dns_ip}:53
+  resolve_retries       10
+  timeout resolve       2s
+  timeout retry         1s
+  hold other           30s
+  hold refused         30s
+  hold nx              30s
+  hold timeout         30s
+  hold valid           30s
+  hold obsolete        30s
```
alkar: I still think we shouldn't veer from the default, especially looking at `obsolete`. If I understand the docs correctly it seems that when an ELB changes IP addresses, the old one will still be valid for an extra 30secs and can potentially cause issues again? (although in my experience, even after the AWS nameserver stops responding with an IP address it will still be valid at least for a few minutes)

haproxy/variables.tf
razvan-moj: Bleh. `Error: variable "dns_ip": default may not contain interpolations`

alkar: Yea you can't do that but you can drop the var and feed it into the template directly ;)

PR #101, opened 2018-11-16
Automate metrics-server install

sablumiah:
WHAT

Use terraform to automate installation of metrics server on k8s clusters

WHY

Able to run kubectl top nodes/pods on cluster

### DIFF
--- a/terraform/cloud-platform-components/metris-server.tf
+++ b/terraform/cloud-platform-components/metris-server.tf
@@ -1,0 +1,30 @@
+resource "helm_release" "metrics_server" {
+  name      = "metrics-server"
+  chart     = "stable/metrics-server"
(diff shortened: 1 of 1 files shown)


terraform/cloud-platform-components/metris-server.tf
```diff
@@ -0,0 +1,29 @@
...
+image:
+  repository: gcr.io/google_containers/metrics-server-amd64
+  tag: v0.3.1
+  pullpolicy: IfNotPresent
+
+args:
+  value[0]: "--kubelet-insecure-tls"
+  value[1]: "--logtostderr"
+
+EOF
```
jasonBirchall: Do we actually need to define a `values` file at all?

Can we just use the default values?

PR #106, opened 2018-11-26
Fluentd/monitoring

ollieanwyll:


helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -56,6 +56,31 @@ roleRef:
   name: {{  template "fluentd-es.fullname" . }}
   apiGroup: ""
 ---
+apiVersion: v1
+kind: Service
+metadata:
+  name: fluentd-es-metrics
+  namespace: logging
```
alkar: Please omit `namespace` as with the other resources. `helm` manages it separately.

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -56,6 +56,31 @@ roleRef:
   name: {{  template "fluentd-es.fullname" . }}
   apiGroup: ""
 ---
+apiVersion: v1
+kind: Service
+metadata:
+  name: fluentd-es-metrics
```
alkar: This should use the template as the other resources do.

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -56,6 +56,31 @@ roleRef:
...
+  namespace: logging
+  labels:
+    k8s-app: fluentd-es
+    version: {{ .Values.image.tag }}
+    kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: Reconcile
```
alkar: I believe `kubernetes.io/cluster-service` and `addonmanager.kubernetes.io/mode` won't do anything (the latter is probably a remnant of the past, the former might be significant for the actual `DaemonSet`). However, I suppose it's fine to leave them in for now - at least it's uniform across all resources.

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -56,6 +56,31 @@ roleRef:
...
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+spec:
+  ports:
+  - name: metrics
+    port: 24231
+    protocol: TCP
+    targetPort: 24231
+  selector:
```
alkar: This works but on a closer read I think I mistakenly picked these two labels. I think the selector here should be the same as the `DaemonSet` selector on L98-101.

I'm not sure which one is better, however, it certainly shouldn't select based on hardcoded values when `app` and `release` are both generated (also, `app` and `release` specifically are equal, see the `DaemonSet` spec below).

Perhaps the following two would work nicely?
```
     app: {{ .Release.Name }}
     version: {{ .Values.image.tag }}
```
This combination seems unique enough.

helm-charts/fluentd-es/templates/fluentd-es.yaml
```diff
@@ -77,8 +76,8 @@ spec:
     protocol: TCP
     targetPort: 24231
   selector:
-    app: fluentd-es
-    release: fluentd-es
+      app: {{ .Release.Name }}
```
alkar: The indentation is wrong here. Can you please fix?

PR #112, opened 2018-11-29
Add labels to kops template & clean-up unused manifests 

jasonBirchall:
**Overview**
---
This PR ensures any resources managed by kops will comply with the labeling standards as outlined here: https://github.com/ministryofjustice/technical-guidance/blob/master/standards/documenting-infrastructure-owners.md. 

It will also clean-up unused manifest files and remove the directory, `/cluster-components`.

**Implementation**
---
Merging this PR will NOT apply the changes. A manual `kops update` or `kops apply` will need to ensue.

PR #113, opened 2018-11-30
Kops template label change

jasonBirchall:
**Overview**
---
Following on from PR #112 the below request was made. This PR encompasses that request along with the removal of the `component` tag, as this was duplicated.

```
stevemarshall [19:06]
@jasonb That’s ace! One minor thing: `business-unit` should be `platforms`. Also, I might argue for `MOJ Cloud Platform` for `application`, and `infrastructure-support` is optional (for cases like this).
So that was three things, actually. Still: thanks for doing this so quickly!
```

### DIFF
--- a/terraform/cloud-platform/templates/kops.yaml.tpl
+++ b/terraform/cloud-platform/templates/kops.yaml.tpl
@@ -313,10 +313,8 @@ spec:
   maxSize: 1
   minSize: 1
   cloudLabels:
-    application: justice-cloud-platform
-    business-unit: hq
-    component: master
-    infrastructure-support: platforms@digital.justice.gov.uk
+    application: moj-cloud-platform
+    business-unit: platforms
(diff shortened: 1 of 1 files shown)


terraform/cloud-platform/templates/kops.yaml.tpl
```diff
@@ -313,10 +313,8 @@ spec:
   maxSize: 1
   minSize: 1
   cloudLabels:
-    application: justice-cloud-platform
-    business-unit: hq
-    component: master
```
SteveMarshall: Did you mean to remove the `component`s? I think it's actually good that we had those…

jasonBirchall: It's a duplicate tag, already defined by kops under `kops.k8s.io/instancegroup`

PR #124, opened 2018-12-17
Introduce cert-manager component

razvan-moj:
**Why**
https://github.com/ministryofjustice/cloud-platform/issues/562


alkar:
I think we should leave this PR open for ministryofjustice/cloud-platform#587 to build on. Merging right now has no value since we cannot apply on live clusters as-is.

alkar:
Squashed, amended and rebased on top of master.

terraform/cloud-platform-components/resources/nginx-ingress/default-certificate.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Certificate
+metadata:
+  name: default
+  namespace: ingress-controllers
+spec:
+  secretName: default-certificate
+  issuerRef:
+    name: letsencrypt-production
+    kind: ClusterIssuer
+  commonName: '*.apps.cloud-platform-test-8.k8s.integration.dsd.io'
```
jasonBirchall: As discussed. Will need to grab this value from upstream tf state.

terraform/cloud-platform-components/resources/nginx-ingress/default-certificate.yaml
```diff
@@ -0,0 +1,17 @@
...
+  acme:
+    config:
+    - dns01:
+        provider: route53
+      domains:
+      - '*.apps.cloud-platform-test-8.k8s.integration.dsd.io'
```
jasonBirchall: Same as above {cloud-platform-test-8.k8s.integration.dsd.io}

alkar:
@jasonBirchall can you apply this latest commit?

PR #125, opened 2018-12-18
Allow prometheus-operator to consume PrometheusRules from all namespaces

alkar:
Connects to: https://github.com/ministryofjustice/cloud-platform/issues/482
Follows up on: https://github.com/ministryofjustice/cloud-platform-environments/pull/234

### DIFF
--- a/terraform/cloud-platform-components/prometheus.tf
+++ b/terraform/cloud-platform-components/prometheus.tf
@@ -15,12 +15,21 @@ resource "kubernetes_storage_class" "prometheus_storage" {
   }
 }
 
+data "template_file" "prometheus_operator" {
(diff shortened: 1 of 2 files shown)


terraform/cloud-platform-components/prometheus.tf
```diff
@@ -15,13 +15,22 @@ resource "kubernetes_storage_class" "prometheus_storage" {
   }
 }
 
+data "template_file" "prometheus_operator" {
+  template = "${file("${path.module}/templates/prometheus-operator.yaml.tpl")}"
+  vars     = {}
+}
+
```
jasonBirchall: As this file doesn't exist, won't it just pass an empty values file?

PR #127, opened 2018-12-28
Kiam cluster component

alkar:
Connects to https://github.com/ministryofjustice/cloud-platform/issues/564

In order to simplify the setup, kiam agent `Pods` run only on node hosts. This allows other `Pods` running on master hosts to assume the host's IAM role as before and avoids breaking any of the kubernetes components. We do not intent to run user workload on master nodes so this should be safe. However, it is probably now worth enabling the [`PodTolerationRestriction` admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podtolerationrestriction) to actually prevent user workload from being scheduled on a master node.

Because the trust (assume) policy uses an instance role as a principal (see the example in the README), we need to run kiam server `Pods` on a specific set of hosts (masters or nodes). This implementation places kiam server `Pods` on node hosts alongside the agents. Alternatively, based on [this document](https://github.com/uswitch/kiam/blob/master/docs/IAM.md) kiam-server can be assigned a dedicated role that it will assume before requesting credentials for any other `Pod`. This has not been implemented here, to reduce complexity.

Although seemingly not critical to the operation of kiam, the latest version of the chart does not include [this fix](https://github.com/uswitch/kiam/pull/174) which causes some error to appear in the kiam-server logs.

### DIFF
--- a/terraform/cloud-platform-components/kiam.tf
+++ b/terraform/cloud-platform-components/kiam.tf
@@ -105,10 +105,11 @@ data "template_file" "kiam" {
 }
 
 resource "helm_release" "kiam" {
-  name      = "kiam"
-  chart     = "stable/kiam"
-  namespace = "kiam"
-  version   = "2.0.0-rc3"
+  name          = "kiam"
+  chart         = "stable/kiam"
+  namespace     = "kiam"
+  version       = "2.0.0-rc3"
+  recreate_pods = "true"
 
   values = [
     "${data.template_file.kiam.rendered}",
--- a/terraform/cloud-platform-components/kiam.tf
+++ b/terraform/cloud-platform-components/kiam.tf
@@ -177,3 +177,14 @@ resource "kubernetes_service" "agent-metrics" {
     }
   }
 }
+
+resource "null_resource" "kiam_servicemonitor" {
+  provisioner "local-exec" {
+    command = "kubectl apply -f ${path.module}/resources/kiam-servicemonitor.yaml"
+  }
+
+  provisioner "local-exec" {
+    when    = "destroy"
+    command = "kubectl delete -f ${path.module}/resources/kiam-servicemonitor.yaml"
+  }
+}
--- a/terraform/cloud-platform-components/kiam.tf
+++ b/terraform/cloud-platform-components/kiam.tf
@@ -129,6 +129,8 @@ resource "helm_release" "kiam" {
 }
 
 resource "kubernetes_service" "server-metrics" {
+  depends_on = ["helm_release.kiam"]
+
   metadata {
     name      = "kiam-server-metrics"
     namespace = "kiam"
@@ -154,6 +156,8 @@ resource "kubernetes_service" "server-metrics" {
 }
 
 resource "kubernetes_service" "agent-metrics" {
+  depends_on = ["helm_release.kiam"]
+
   metadata {
     name      = "kiam-agent-metrics"
     namespace = "kiam"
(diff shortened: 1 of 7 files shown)


terraform/cloud-platform-components/kiam.tf
```diff
@@ -0,0 +1,190 @@
...
+  }
+}
+
+resource "tls_locally_signed_cert" "agent" {
+  cert_request_pem   = "${tls_cert_request.agent.cert_request_pem }"
+  ca_key_algorithm   = "${tls_private_key.ca.algorithm}"
+  ca_private_key_pem = "${tls_private_key.ca.private_key_pem}"
+  ca_cert_pem        = "${tls_self_signed_cert.ca.cert_pem}"
+
+  validity_period_hours = 8760 // 1 year
```
jasonBirchall: We need a way to manage the expiry of these certificates. 
Will kiam just stop working? 
Is it worth pattern matching a triggered alert?
When it expires, is it a case of re-running this Terraform file?

terraform/cloud-platform/templates/kops.yaml.tpl
```diff
@@ -170,29 +170,6 @@ spec:
...
-          "Effect": "Allow",
-          "Action": [
-            "route53:ListHostedZones",
-            "route53:ListResourceRecordSets"
-          ],
-          "Resource": [
-            "*"
-          ]
-        }
-      ]
```
jasonBirchall: Built a test cluster using this branch and the above changes. Unfortunately, `kops` doesn't ever validate. 
I just end up with:
```
Using cluster from kubectl context: cloud-platform-kiam-test.k8s.integration.dsd.io

Validating cluster cloud-platform-kiam-test.k8s.integration.dsd.io

unexpected error during validation: unable to resolve Kubernetes cluster API URL dns: lookup api.cloud-platform-kiam-test.k8s.integration.dsd.io on xxx.xxx.x.x(redacted):53: no such host
```

jasonBirchall: I may have been a little haste. :)

This did eventually validate. Nothing to see here.

terraform/cloud-platform-components/kiam.tf
```diff
@@ -0,0 +1,190 @@
...
+  metadata {
+    name      = "kiam-server-metrics"
+    namespace = "kiam"
+
+    labels {
+      app       = "kiam"
+      component = "server-metrics"
+    }
+  }
+
```
jasonBirchall: On fresh clusters the `kiam` namespace won't exist and the service creation will fail. Perhaps put a ns creation before.

jasonBirchall: or put a dependency allowing the Helm to run first (and create the ns if it doesn't exist). 

This is the error I'm seeing. Easily remedied by creating the ns manually:
`* kubernetes_service.agent-metrics: namespaces "kiam" not found`

alkar: @jasonBirchall e9a4931 should fix this but I have not tested it yet

terraform/cloud-platform-components/kiam.tf
alkar: As discussed:
- `kiam` will just stop working, indeed but with it, any `Pods` using IAM roles.
- We could use `blackbox_exporter` to get cert expiry in a prom metric and hook an alert on it. However, we will most likely drive this off https://github.com/ministryofjustice/cloud-platform/issues/587 so it's all done (hopefully) transparently. Should we just create a calendar event for now and raise an issue to address?
- Yes, re-running tf will renew & fix it.

terraform/cloud-platform-components/kiam.tf
alkar: This has now been tested.

terraform/cloud-platform-components/kiam.tf
jasonBirchall: Yes, although not ideal the calendar entry will suffice.

PR #128, opened 2019-01-09
cleaned up tests

razvan-moj:


kops/cloud-platform-test-7.yaml
```diff
@@ -174,7 +179,7 @@ spec:
             "route53:ChangeResourceRecordSets"
           ],
           "Resource": [
```
jasonBirchall: Not sure you'll want to include this part. It's being taking out as of https://github.com/ministryofjustice/cloud-platform-infrastructure/pull/127/files#diff-ad3d1a514b5db6bf91fbc0285d1e920cL173

PR #146, opened 2019-02-04
Add AM alerting routes documentation

sablumiah:
WHAT
Need a way to route specific alerts to named slack channels 

WHY 
Custom alerts should only notify the teams/channels that want to know about it

### DIFF
--- a/terraform/cloud-platform-components/README.md
+++ b/terraform/cloud-platform-components/README.md
@@ -41,3 +41,123 @@ resource "aws_iam_role_policy" "app" {
 ```
 
 This can easily be configured as part of a user environment's resources, along with the required namespace annotation (see the [kiam docs](https://github.com/uswitch/kiam#overview)).
+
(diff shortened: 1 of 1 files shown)


terraform/cloud-platform-components/README.md
```diff
@@ -41,3 +41,123 @@ resource "aws_iam_role_policy" "app" {
...
+```
+#### Add a new variable in `variables.tf`
+
+```bash
+variable "slack_config_<teamn_name>" {
+  description = "Add Slack webhook API URL and channel for integration with slack."
+}
+```
+
+All alerts are routed using the `severity` label. Provide the development team the severity label created for each route (default is team_name),
```
alkar: Can we rename the label now before it's much more hassle to do so (perhaps `team` or `receiver` or something else)? Perhaps it doesn't matter that much?

Should we also rename ours to something like `cloud-platform-critical` and `cloud-platform-warning` ?

PR #151, opened 2019-02-13
Specify role for external-dns

alkar:
Fixes an issue where external-dns assumes the node's role which has no access to edit the route53 zone.


alkar:
@jasonBirchall pushed a missing resource, can you check again please? tyvm!

PR #156, opened 2019-02-18
add inactive user check template

sid234:
To increase the security of the AWS account, we need to remove IAM user credentials (that is, passwords and access keys) that are not needed.
The template deactivates unused credentials and then moves the user accounts to the suspended user group with deny all policy


sid234:
updated with cfn-cli-package

PR #181, opened 2019-03-01
Add/fluentd buffer alarm

ollieanwyll:
This PR connect to ministryofjustice/cloud-platform#673

If merged this PR will add a Prometheus alarm to the Helm chart of Fluentd.

The alarm is triggered when the Fluentd buffer size is nearly full for over 2 minutes.

The metric is measured in total byte size. The actual buffer limit is 256M, but this alarm is trigger at anything over 180M.

### DIFF
--- a/helm-charts/fluentd-es/templates/prometheus-custom-rules-fluentd.yaml
+++ b/helm-charts/fluentd-es/templates/prometheus-custom-rules-fluentd.yaml
@@ -1,19 +1,19 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
-  namespace: <namespace>
+  namespace: logging
   labels:
     role: alert-rules
-  name: prometheus-custom-rules-<application_name>
+  name: prometheus-custom-rules-fluentd
 spec:
   groups:
   - name: application-rules
     rules:
-    - alert: <alert_name>
-      expr: <alert_query>
-      for: <check_time_length>
+    - alert: fluentd_buffer_full
+      expr: fluentd_output_status_buffer_total_bytes > 1800000000
+      for: 2m
       labels:
-        severity: <team_name>
+        severity: cp-team
       annotations:
-        message: <alert_message> 
-        runbook_url: <http://my-support-docs>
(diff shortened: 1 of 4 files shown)


helm-charts/fluentd-es/templates/prometheus-custom-rules-fluentd.yaml
```diff
@@ -1,19 +1,19 @@
...
+      expr: fluentd_output_status_buffer_total_bytes > 1800000000
+      for: 2m
       labels:
-        severity: <team_name>
+        severity: cp-team
       annotations:
-        message: <alert_message> 
-        runbook_url: <http://my-support-docs>
\ No newline at end of file
+        message: The Fluentd buffer limit of 8 x 256Mb chunks is full. This indicates that Fluentd is having issues writing to the ElasticSearch cluster or it is collecting logs quicker than it can write.
```
digitalronin: Is there any way to pull the values from `helm-charts/fluentd-es/config/output.conf` and use them here? 

Assuming we can't do that, I'd suggest changing this to; "The Fluentd buffer limit (defined by the `chunk_limit_size` and `queue_limit_length` values in `helm-charts/fluentd-es/config/output.conf`) is full..."

That way, the error message will still be accurate if those values are changed.

ollieanwyll: I've had a quick chat with Mitch and we both don't think it would be possible to pull the values.

But I have updated the message. :)

digitalronin: 👍 I thought that might be tricky.

alkar:
Do we want to make the `PrometheusRule` part of the helm chart? This creates a hard dependency on the prometheus operator.

If we *do* want to do this, I think we should also pull the [relevant `ServiceMonitor`](https://github.com/ministryofjustice/cloud-platform-environments/blob/master/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/logging/monitoring.yaml) in the chart as well.

PR #191, opened 2019-03-11
feature/account-baseline-templates

sid234:
aws-config-cloudtrail-logging templates enable config service and adds basic config rules, enable cloudtrail service and creates s3 bucket to store cloudtrail logs
aws-suspend-inactive-users template checks for inactive iam users and deactivates their login profile and access keys
aws-iam-user-roles-groups has set of templates that creates user assume role permissions, instance profiles, user mfa enforcement role and policy
aws-s3-object-auto-remediation template checks reactive approach when s3 object is made public, the function updates the permissions to private
the readme document provides background, instructions to deploy and references
aws-risk-credentials-exposed has template that creates the functions and related resources to check for exposed credentials or access keys from the account


razvan-moj:
- aws-config-enable-rules fails with `The runtime parameter of nodejs4.3 is no longer supported for creating or updating AWS Lambda functions. We recommend you use the new runtime (nodejs8.10) while creating or updating functions. 
 - aws-risk-credentials-serverless fails with `Resource with id [DeleteAccessKeyPair] is invalid.`

PR #200, opened 2019-03-13
Changes Grafana auth method from GitHub to Auth0

ollieanwyll:
Connects to issue ministryofjustice/cloudplatform#642

This PR if merged with change the authentication method used for Grafana from the GitHub provided 0Auth to Auth0.

This aligns the authentication method with the rest of our cluster hosted services.

This is for the live-0 cluster.


terraform/cloud-platform-components/templates/prometheus-operator.yaml.tpl
```diff
@@ -465,10 +465,13 @@ grafana:
     GF_USERS_ALLOW_SIGN_UP: "false"
     GF_USERS_AUTO_ASSIGN_ORG_ROLE: "Viewer"
     GF_USERS_VIEWERS_CAN_EDIT: "true"
-    GF_AUTH_GITHUB_ENABLED: "true"
-    GF_AUTH_GITHUB_ALLOW_SIGN_UP: "true"
     GF_SMTP_ENABLED: "false"
-    GF_AUTH_GITHUB_ALLOWED_ORGANIZATIONS: "ministryofjustice"
+    GF_AUTH_GENERIC_OAUTH_ENABLED: "true"
+    GF_AUTH_GENERIC_OAUTH_ALLOW_SIGN_UP: "true"
+    GF_AUTH_GENERIC_OAUTH_TEAM_IDS:
```
alkar: This and the following value are empty, do we need them here?

terraform/cloud-platform-components/prometheus.tf
```diff
@@ -8,9 +8,11 @@ resource "kubernetes_secret" "grafana_secret" {
   }
 
   data {
-    GF_AUTH_GITHUB_CLIENT_ID     = "${ var.github_client_id }"
-    GF_AUTH_GITHUB_CLIENT_SECRET = "${ var.github_client_secret }"
-    GF_SECURITY_SECRET_KEY       = "${ var.github_secret_key }"
+    GF_AUTH_GENERIC_OAUTH_CLIENT_ID     = "${ var.auth0_client_id }"
```
alkar: These variables should be picked up from the `cloud-platform` [state](https://github.com/ministryofjustice/cloud-platform-infrastructure/blob/live-0/terraform/cloud-platform/outputs.tf#L53-L71). There is an auth0 client created there to be used for components.

Having them as variables in this terraform configuration means we need to manually create the app for every cluster and also manually pass them in every time we need to run terraform, which is something we should avoid.

terraform/cloud-platform-components/templates/prometheus-operator.yaml.tpl
ollieanwyll: Looking at what other people have done, they have let them blank.
I will test it without the values.

PR #202, opened 2019-03-13
new alert route in alertmanager laa-apply-for-legal-aid live-0

pwyborn:
What:
See ticket https://waffle.io/ministryofjustice/cloud-platform/cards/5c8777ee7a4a6a0039ba24e8
New alert manager route


terraform/cloud-platform-components/templates/prometheus-operator.yaml.tpl
```diff
@@ -452,11 +515,9 @@ grafana:
     ## Secret must be manually created in the namespace
     ##
     tls:
-      - hosts:
-        - "${ grafana_ingress }"
-    # - secretName: prometheus-general-tls
-    #   hosts:
-    #   - prometheus.example.com
+    - secretName: prometheus-general-tls
```
alkar: Does `secretName` need to be uncommented?
It was left as a comment so that the Grafana Ingress would use the default wildcard certificate. `prometheus-general-tls` is not being created anywhere?

pwyborn:
Hi @alkar - made changes - hopefully have got it right now?

terraform/cloud-platform-components/templates/prometheus-operator.yaml.tpl
pwyborn: Hi @alkar - made changes - hopefully have got it right now?

alkar:
Hi @pwyborn I think I didn't explain properly. We should keep the `tls` block but not define a `secretName` in it. I've pushed two commits (see the last two) in order to merge this because I need to run some operations on the cluster. Happy to chat about it offline!

PR #228, opened 2019-03-26
Revisit the auth0 oidc group script

alkar:
This should now take into account only teams within the ministryofjustice org.


kerin:
@alkar FYI, analytics has [this rule](https://github.com/ministryofjustice/analytics-platform-auth0/blob/alpha/rules/whitelist-github-orgs.js), which runs first and bails out with an authZ error if the user isn't a member of an allowed org.

PR #232, opened 2019-03-28
Custom Alert Rules added to operator install - live0

sablumiah:
WHAT
Add custom alerts to Prometheus and apply rules as part of the prometheus-operator install

WHY
The custom rules monitor CP team namespaces as well as system namespaces and are based on the business unit-cloud-platform annotation. The auto apply of these rules also mean they will not be missed if there is a need to reinstall prometheus-operator. Associated runbbook does have also been updated with runbook_url added to all alerts.

### DIFF
--- a/terraform/cloud-platform-components/prometheus.tf
+++ b/terraform/cloud-platform-components/prometheus.tf
@@ -61,6 +61,10 @@ resource "helm_release" "prometheus_operator" {
     "null_resource.deploy",
   ]
 
+  provisioner "local-exec" {
+    command = "kubectl apply -n monitoring -f ${path.module}/resources/prometheusrule-alerts/"
+  }
+
   # Delete Prometheus leftovers
   # Ref: https://github.com/coreos/prometheus-operator#removal
   provisioner "local-exec" {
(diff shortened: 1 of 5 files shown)


terraform/cloud-platform-components/prometheus.tf
```diff
@@ -61,6 +61,10 @@ resource "helm_release" "prometheus_operator" {
     "null_resource.deploy",
   ]
 
+  provisioner "local-exec" {
+    command = "kubectl apply -n monitoring -f ${path.module}/resources/prometheusrule-alerts/"
```
jasonBirchall: You'll need to define a local-exec with the delete command.

sablumiah: I did not add it, as part of the uninstall of operator all rules get deleted including all the custom rules, so when the delete command runs, it does not find anything to delete

alkar: I think we should add it then with the `--ignore-not-found` flag. So when we fix that CRD issue (which might have been addressed upstream) it just works properly.

Additionally, should this be a separate `null_resource` instead? I think it would improve lifecycle management (you can `taint / apply / destroy` independently from the prom chart). Is there a reason why this should be different to the proxy, for example?

PR #235, opened 2019-03-29
Shorter hostnames for live ingresses

razvan-moj:
**Why**
We don't want the main monitoring/login page address to change when we switch clusters

### DIFF
--- a/terraform/cloud-platform-components/external-dns.tf
+++ b/terraform/cloud-platform-components/external-dns.tf
@@ -16,11 +16,12 @@ resource "aws_iam_role" "external_dns" {
 
 data "aws_iam_policy_document" "external_dns" {
   statement {
-    actions   = ["route53:ChangeResourceRecordSets"]
+    actions = ["route53:ChangeResourceRecordSets"]
+
     resources = [
       "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}",
-      "${terraform.workspace == local.live_workspace ? format("%s/%s", "arn:aws:route53:::hostedzone", data.terraform_remote_state.global.cp_zone_id) : "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}"}"
-      ]
+      "${terraform.workspace == local.live_workspace ? format("%s/%s", "arn:aws:route53:::hostedzone", data.terraform_remote_state.global.cp_zone_id) : "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}"}",
+    ]
   }
 
   statement {
--- a/terraform/cloud-platform-components/external-dns.tf
+++ b/terraform/cloud-platform-components/external-dns.tf
@@ -17,7 +17,10 @@ resource "aws_iam_role" "external_dns" {
 data "aws_iam_policy_document" "external_dns" {
   statement {
     actions   = ["route53:ChangeResourceRecordSets"]
-    resources = ["arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}"]
+    resources = [
+      "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}", 
+      "arn:aws:route53:::hostedzone/${data.terraform_remote_state.global.justice_gov_uk.zone_id}"
+      ]
   }
 
   statement {
--- a/terraform/cloud-platform-components/external-dns.tf
+++ b/terraform/cloud-platform-components/external-dns.tf
@@ -18,10 +18,10 @@ data "aws_iam_policy_document" "external_dns" {
   statement {
     actions = ["route53:ChangeResourceRecordSets"]
 
-    resources = [
+    resources = ["${compact(list(
       "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}",
-      "${terraform.workspace == local.live_workspace ? format("%s/%s", "arn:aws:route53:::hostedzone", data.terraform_remote_state.global.cp_zone_id) : "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}"}",
-    ]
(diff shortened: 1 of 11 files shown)


terraform/cloud-platform-components/templates/oauth2-proxy.yaml.tpl
```diff
@@ -29,7 +29,7 @@ ingress:
   enabled: true
   path: /
   hosts:
-    - "${hostname}"
+    - "${hostname}"
```
jasonBirchall: Remove space.

jasonBirchall: Done in cb4f98b.

terraform/cloud-platform-components/nginx-ingress-acme.tf
```diff
@@ -30,7 +30,7 @@ controller:
 
   service:
     annotations:
-      external-dns.alpha.kubernetes.io/hostname: "*.apps.${data.terraform_remote_state.cluster.cluster_domain_name},apps.${data.terraform_remote_state.cluster.cluster_domain_name}"
+      external-dns.alpha.kubernetes.io/hostname: "${terraform.workspace == local.live_workspace ? format("*.%s, *.apps.%s, apps.%s", local.live_domain, data.terraform_remote_state.cluster.cluster_domain_name, data.terraform_remote_state.cluster.cluster_domain_name) : format("*.apps.%s, apps.%s", data.terraform_remote_state.cluster.cluster_domain_name, data.terraform_remote_state.cluster.cluster_domain_name)}"
```
alkar: I find this a bit hard to parse, can't we instead do something like:
```
"*.apps.${data.terraform_remote_state.cluster.cluster_domain_name},apps.${data.terraform_remote_state.cluster.cluster_domain_name}${terraform.workspace == local.live_workspace ? format(",*.%s", local.live_domain) : ""}"
```

terraform/cloud-platform-components/nginx-ingress-acme.tf
```diff
@@ -59,6 +59,7 @@ data "template_file" "nginx_ingress_default_certificate" {
 
   vars {
     common_name = "*.apps.${data.terraform_remote_state.cluster.cluster_domain_name}"
+    alt_name    = "${terraform.workspace == local.live_workspace ? format("- *.%s", local.live_domain) : ""}"
```
alkar: Is that `-` correct?

terraform/cloud-platform-components/external-dns.tf
```diff
@@ -16,8 +16,12 @@ resource "aws_iam_role" "external_dns" {
 
 data "aws_iam_policy_document" "external_dns" {
   statement {
-    actions   = ["route53:ChangeResourceRecordSets"]
-    resources = ["arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}"]
+    actions = ["route53:ChangeResourceRecordSets"]
+
+    resources = [
+      "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}",
```
alkar: When that ternary is false, this will contain the same zone id twice. Can we not use an empty string for false?

terraform/cloud-platform-components/fluentd.tf
```diff
@@ -5,12 +5,12 @@ resource "helm_release" "fluentd_es" {
 
   set {
     name  = "fluent_elasticsearch_host"
-    value = "${terraform.workspace == local.live_workspace ? "search-cloud-platform-live-dibidbfud3uww3lpxnhj2jdws4.eu-west-2.es.amazonaws.com" : "search-test-yuq2c7ziybjvpmr2tllkghh6va.eu-west-2.es.amazonaws.com"}"
+    value = "${replace(terraform.workspace, "live", "") != terraform.workspace ? "search-cloud-platform-live-dibidbfud3uww3lpxnhj2jdws4.eu-west-2.es.amazonaws.com" : "search-test-yuq2c7ziybjvpmr2tllkghh6va.eu-west-2.es.amazonaws.com"}"
```
alkar: Again with this, I think it's a bit hard to parse. Do we currently need to plan for other live clusters? I don't believe we plan to get any more anytime soon. Instead, the previous statement was easier to understand. Wdyt?

terraform/cloud-platform-components/nginx-ingress-acme.tf
razvan-moj: yes, sadly, indenting :(

terraform/cloud-platform-components/external-dns.tf
razvan-moj: couldn't find a way, terraform insist on sending an empty pair of quotes that breaks indenting

terraform/cloud-platform-components/fluentd.tf
razvan-moj: we have live-0 and live-1 so

terraform/cloud-platform-components/nginx-ingress-acme.tf
razvan-moj: neat! changed

terraform/cloud-platform-components/fluentd.tf
alkar: I'm saying this because `live-0` is not managed from this branch and won't live long enough to try and integrate. Is `live-2` even a concern right now that would warrant complicating this?

terraform/cloud-platform-components/nginx-ingress-acme.tf
alkar: I think you missed committing changes to the cert template?

terraform/cloud-platform-components/external-dns.tf
alkar: Do you mean it breaks validation with the AWS API?
We could do something like this then:
```
["${compact(list(
      "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}",
      "${terraform.workspace == local.live_workspace ? format("%s/%s", "arn:aws:route53:::hostedzone", data.terraform_remote_state.global.cp_zone_id) : ""}",
    ))}"]
```
But I'm not sure this reads better 🍝 
It's just that the fact that the duplicates get pruned when you actually apply is perhaps a "feature" we shouldn't depend upon (couldn't find any mention of this) 🤷‍♂️

PR #244, opened 2019-04-10
Add runbook for rotating git-crypt key

alkar:
Documents the process of rotating the `git-crypt` encryption key.

### DIFF
--- a/docs/rotate-git-crypt.md
+++ b/docs/rotate-git-crypt.md
@@ -8,12 +8,16 @@ When rotating the `git-crypt` symmetric key, you should follow the steps below:
 
 2. Make a list of the current users:
   ```
-  ls .git-crypt/keys/default/0/*.gpg | sed -E 's/^.+\/([A-Z0-9]{40}).gpg$/\1/' > git-crypt-users
+  # for every file found, extract its filename minus the suffix,
+  # which corresponds to the user's key id
+  ls .git-crypt/keys/default/0/*.gpg | xargs -I{} -- basename {} .gpg > git-crypt-users
   ```
 
 2. Make a list of the encrypted files:
   ```
-  git-crypt status -e | sed -E 's/^.+encrypted:[[:space:]]+([^[:space:]]+)$/\1/' > git-crypt-files
+  # for every currently encrypted file, extract its path, relative
+  # to the root of the repository
+  git-crypt status -e | awk '{ print $2; }' > git-crypt-files
   ```
 
 3. Remove `git-crypt` configuration:
@@ -38,7 +42,7 @@ When rotating the `git-crypt` symmetric key, you should follow the steps below:
 5. Re-encrypt all the secrets: the list of secrets was extracted in `git-crypt-files` (see step 2). Since the master encryption key is being rotated, all of the secrets that have been encrypted with it *must* be rotated. Once finished, you can add the files to the index
 
   ```
-  cat git-crypt-files | xargs -I{} git add
+  cat git-crypt-files | xargs git add
   ```
 
 6. Commit your changes. One way to check that the files in the git index are properly encrypted before you commit your changes is like so:
--- a/docs/rotate-git-crypt.md
+++ b/docs/rotate-git-crypt.md
@@ -1,0 +1,55 @@
+# Rotating the git-crypt key
+
+We use `git-crypt` for securing secrets committed to `git`. It uses a symmetric encryption key which is then encrypted for users using their GPG keys.
+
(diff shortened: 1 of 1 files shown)


docs/rotate-git-crypt.md
```diff
@@ -0,0 +1,55 @@
...
+
+2. Make a list of the current users:
+  ```
+  ls .git-crypt/keys/default/0/*.gpg | sed -E 's/^.+\/([A-Z0-9]{40}).gpg$/\1/' > git-crypt-users
+  ```
+
+2. Make a list of the encrypted files:
+  ```
+  git-crypt status -e | sed -E 's/^.+encrypted:[[:space:]]+([^[:space:]]+)$/\1/' > git-crypt-files
+  ```
```
digitalronin: Couldn't this be simpler?

```
git-crypt status -e | sed 's/.*encrypted: //' > git-crypt-files
```

docs/rotate-git-crypt.md
```diff
@@ -0,0 +1,55 @@
...
+We use `git-crypt` for securing secrets committed to `git`. It uses a symmetric encryption key which is then encrypted for users using their GPG keys.
+
+When rotating the `git-crypt` symmetric key, you should follow the steps below:
+
+1. Ensure you have unlocked the repository: `git-crypt unlock`
+
+2. Make a list of the current users:
+  ```
+  ls .git-crypt/keys/default/0/*.gpg | sed -E 's/^.+\/([A-Z0-9]{40}).gpg$/\1/' > git-crypt-users
+  ```
```
digitalronin: This is quite hard to read. How about this?

```
ls .git-crypt/keys/default/0/*.gpg \
  | sed 's/.*default.0.//' \
  | sed 's/.gpg//' \
  > git-crypt-users
```

docs/rotate-git-crypt.md
```diff
@@ -0,0 +1,55 @@
...
+  git-crypt init
+  ```
+  make sure you have all the users' keys,
+  ```
+  cat git-crypt-users | xargs -I{} gpg --recv-key {}
+  ```
+  and add all the users back,
+  ```
+  cat git-crypt-users | xargs -I{} git-crypt add-gpg-user -n --trusted {}
+  ```
```
digitalronin: 👍 Bonus points for this use of xargs! One of my favourite and most overlooked unix tools.

docs/rotate-git-crypt.md
```diff
@@ -0,0 +1,55 @@
...
+  At this point, your git index will have a number of pending changes, do not commit them yet.
+
+5. Re-encrypt all the secrets: the list of secrets was extracted in `git-crypt-files` (see step 2). Since the master encryption key is being rotated, all of the secrets that have been encrypted with it *must* be rotated. Once finished, you can add the files to the index
+
+  ```
+  cat git-crypt-files | xargs -I{} git add
```
digitalronin: You don't need `-I{}` here, do you?

docs/rotate-git-crypt.md
```diff
@@ -0,0 +1,55 @@
...
+6. Commit your changes. One way to check that the files in the git index are properly encrypted before you commit your changes is like so:
+  ```
+  git show :<path-to-file>
+  ```
+  or after committing (and before you've pushed):
+  ```
+  git show HEAD:<path-to-file>
+  ```
+  where `<path-to-file>` is either absolute from the base of the git repo or relative (eg.: `git show:./my-secret-file.yaml`)
+
```
digitalronin: Please add a note here explaining what the user should see if the file is encrypted correctly (i.e. something like `^@GITCRYPT^@<E5>ݎ^]xxY<F0><8F>PR<BE>^U/"i5i2t^Ep^R<BB><C3><E8>B<EE><A5>bK<85><C5>^B^B<C5>@5<D8><FB>n<9F><DA>w<B3>O<D1><FF>`) instead of the cleartext.

Without that, if that output of the `git show` command is:

```
I AM SOOPER SEKRIT AND SHOULD NOT BE PUSHED TO GITHUB
```

...how does the user know this isn't OK? (admittedly, they should be able to figure it out from context, but it never hurts to be explicit).

docs/rotate-git-crypt.md
alkar: I don't think so. Just a habit because osx `xargs` works somewhat differently.

docs/rotate-git-crypt.md
alkar: As discussed, I've ended up replacing the regular expressions with simpler commands and comments for the reader. It's hopefully easier to read now.

PR #246, opened 2019-04-11
Update external-dns and cert-manager configuration

vijay-veeranki-moj:
This PR is to update external-dns and cert-manager configuration, so that:

-components in live-1 can access all the Route53 zones in the cloud-platform account 
-components in other clusters are only able to access their own dedicated zone

#802

### DIFF
--- a/terraform/cloud-platform-components/cert-manager.tf
+++ b/terraform/cloud-platform-components/cert-manager.tf
@@ -84,7 +84,7 @@ ingressShim:
   defaultIssuerName: letsencrypt-production
   defaultIssuerKind: ClusterIssuer
   defaultACMEChallengeType: dns01
-  defaultACMEDNS01ChallengeProvider: route53
+  defaultACMEDNS01ChallengeProvider: route53-cloud-platform
 
 securityContext:
   enabled: false
--- a/terraform/cloud-platform-components/cert-manager.tf
+++ b/terraform/cloud-platform-components/cert-manager.tf
@@ -23,9 +23,8 @@ data "aws_iam_policy_document" "cert_manager" {
     actions = ["route53:ChangeResourceRecordSets"]
 
     resources = ["${compact(list(
-      "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}",
-      "${terraform.workspace == local.live_workspace ? format("%s/%s", "arn:aws:route53:::hostedzone", data.terraform_remote_state.global.cp_zone_id) : ""}",
-    ))}"]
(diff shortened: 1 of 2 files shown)


terraform/cloud-platform-components/cert-manager.tf
```diff
@@ -23,9 +23,8 @@ data "aws_iam_policy_document" "cert_manager" {
     actions = ["route53:ChangeResourceRecordSets"]
 
     resources = ["${compact(list(
-      "arn:aws:route53:::hostedzone/${data.terraform_remote_state.cluster.hosted_zone_id}",
-      "${terraform.workspace == local.live_workspace ? format("%s/%s", "arn:aws:route53:::hostedzone", data.terraform_remote_state.global.cp_zone_id) : ""}",
-    ))}"]
+      "${terraform.workspace == local.live_workspace ? "arn:aws:route53:::hostedzone/*" : format("%s/%s", "arn:aws:route53:::hostedzone", data.terraform_remote_state.cluster.hosted_zone_id)}",
```
alkar: What if we re-wrote this like so:
`"${format("arn:aws:route53:::hostedzone/%s", terraform.workspace == local.live_workspace ? "*" : data.terraform_remote_state.cluster.hosted_zone_id}"`
I think it's much easier to read? (haven't tested syntax validity!)

terraform/cloud-platform-components/external-dns.tf
```diff
@@ -19,9 +19,8 @@ data "aws_iam_policy_document" "external_dns" {
     actions = ["route53:ChangeResourceRecordSets"]
 
     resources = ["${compact(list(
```
alkar: Same as above.

terraform/cloud-platform-components/cert-manager.tf
```diff
@@ -23,9 +23,8 @@ data "aws_iam_policy_document" "cert_manager" {
     actions = ["route53:ChangeResourceRecordSets"]
 
     resources = ["${compact(list(
```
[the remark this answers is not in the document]
vijay-veeranki-moj: Removed the compact(list())

terraform/cloud-platform-components/cert-manager.tf
vijay-veeranki-moj: updated the "aws_iam_role" "external_dns" resources filter as suggested

terraform/cloud-platform-components/external-dns.tf
vijay-veeranki-moj: Removed the compact(list())

PR #253, opened 2019-04-15
fix template

sid234:
fix commands in readme, sns topic, add s3 exception to public access block, lambda


sid234:
thank you very much

PR #254, opened 2019-04-15
 PSP RoleBinding resource created for OPA

vijay-veeranki-moj:
Being allowed to use the _privileged_ policy means binding it to the default service account of a namespace. A simple RoleBinding resource needs to be created.

### DIFF
--- a/terraform/cloud-platform-components/resources/opa/opa-psp-rbac.yaml
+++ b/terraform/cloud-platform-components/resources/opa/opa-psp-rbac.yaml
@@ -1,0 +1,16 @@
+# Being allowed to use the _privileged_ policy means binding it to the default serviceaccount of a namespace.
+# A simple RoleBinding resource needs to be created, as described below :
+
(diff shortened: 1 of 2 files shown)


terraform/cloud-platform-components/resources/opa/opa-psp-rbac.yaml
```diff
@@ -0,0 +1,16 @@
+# Being allowed to use the _privileged_ policy means binding it to the default serviceaccount of a namespace.
+# A simple RoleBinding resource needs to be created, as described below :
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: PrivilegedRoleBinding
```
alkar: Can we please rename this to something a bit more descriptive? For example `opa:psp-privileged` or similar

vijay-veeranki-moj: Updated name to opa:psp-privileged

PR #263, opened 2019-05-15
create test ES cluster

razvan-moj:
**Why**
A smaller cluster where test-* k8s fluentd will send logs; can have its data wiped at any time


razvan-moj:
> My only issue with this is that it'll fill up very quickly depending on how many people have test Kubernetes clusters (normally one per person during any given sprint).
When loading logs, I just use a very simple es-curator.py to cleanup everything:

PR #265, opened 2019-05-15
Make audit logging optional in the fluentd-es chart

alkar:
This allows us to disable audit log scraping by fluentd, which is very useful for test clusters. The current solution
does not work properly since it's set to log both into the same ElasticSearch cluster, something that's causing issues
due to the difference in mappings and structure of logs.

### DIFF
--- a/terraform/cloud-platform-components/fluentd.tf
+++ b/terraform/cloud-platform-components/fluentd.tf
@@ -10,7 +10,7 @@ resource "helm_release" "fluentd_es" {
 
   set {
     name  = "fluent_elasticsearch_audit_host"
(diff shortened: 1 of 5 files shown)


terraform/cloud-platform-components/fluentd.tf
```diff
@@ -10,7 +10,7 @@ resource "helm_release" "fluentd_es" {
 
   set {
     name  = "fluent_elasticsearch_audit_host"
-    value = "${replace(terraform.workspace, "live", "") != terraform.workspace ? "search-cloud-platform-audit-dq5bdnjokj4yt7qozshmifug6e.eu-west-2.es.amazonaws.com" : "search-test-yuq2c7ziybjvpmr2tllkghh6va.eu-west-2.es.amazonaws.com"}"
+    value = "${replace(terraform.workspace, "live", "") != terraform.workspace ? "search-cloud-platform-audit-dq5bdnjokj4yt7qozshmifug6e.eu-west-2.es.amazonaws.com" : ""}"
```
razvan-moj: cluster address changed, is now search-cloud-platform-test-zradqd7twglkaydvgwhpuypzy4.eu-west-2.es.amazonaws.com

PR #281, opened 2019-06-06
Add Certificate expiry alert

alkar:


terraform/cloud-platform-components/resources/cert-manager/alerts.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - name: cert-manager
+    rules:
+    - alert: CertificateExpiresSoon
+      # cert-manager will renew certificates when they have 30 days left by default,
+      # as per ACME recommendation. When a certificate has less than 25 days left,
+      # this indicates a problem.
+      expr: certmanager_certificate_expiration_timestamp_seconds - time() < 25 * 24 * 3600
+      for: 1d
+      labels:
+        severity: low-priority
```
jasonBirchall: You'll need to use `warning` here. As per https://github.com/ministryofjustice/cloud-platform-infrastructure/blob/master/terraform/cloud-platform-components/templates/prometheus-operator.yaml.tpl#L116

PR #283, opened 2019-06-10
Cloudwatch Event for DLM and Slack Notify Module

sablumiah:
WHAT
Create a cloudwatch event to monitor dlm status and install slack notify module to alert in slack 

WHY
cloudwatch event will monitor dlm as its used to backup all k8s node EBS volumes and we need to know if this backup ever fails.


sablumiah:
dup

PR #293, opened 2019-06-20
Add a script to build a cluster non-interactively

digitalronin:
The process for building a new cluster, as documented in the current runbooks site, has several manual steps.

This PR adds a `create-cluster.sh` script which:

* Checks all pre-requisites (env. vars, software tools installed, aws profiles)
* Validates that the cluster name is acceptable (if too long, things break)
* Builds the cluster infrastructure (VMs, networks, etc.)
* Uses kops to create a kubernetes cluster
* Installs the cluster 'components'

There is also an 'example.env.create-cluster' which shows how to create
a `.env` file containing all the environment variables which the script
needs.

At the time of writing, the script takes just under 30 minutes to build a new
cluster from scratch.

### DIFF
--- a/create-cluster.sh
+++ b/create-cluster.sh
@@ -45,7 +45,7 @@ run_kops() {
   kops create -f kops/${cluster_name}.yaml
 
   # This is a throwaway SSH key which we never need again.
-  rm -rf /tmp/${cluster_name}*
+  rm -f /tmp/${cluster_name} /tmp/${cluster_name}.pub
   ssh-keygen -b 4096 -P '' -f /tmp/${cluster_name}
 
   kops create secret --name ${cluster_name}.${CLUSTER_SUFFIX} sshpublickey admin -i /tmp/${cluster_name}.pub
--- a/create-cluster.sh
+++ b/create-cluster.sh
@@ -1,0 +1,177 @@
+#!/bin/bash
+
+# See example.env.create-cluster for the environment variables
+# which must be set before running this script.
+#
+# Usage:
+#
+#   ./create-cluster.sh [cluster-name]
+#
+
+readonly MAX_CLUSTER_NAME_LENGTH=12
+readonly CLUSTER_SUFFIX=cloud-platform.service.justice.gov.uk
+# TODO: use the right dns flush command, depending on the architecture of the local machine
+readonly DNS_FLUSH_COMMAND='sudo killall -HUP mDNSResponder' # Mac OSX Mojave
+
+set -euo pipefail
+
+main() {
+  local readonly cluster_name=$1
+
+  check_prerequisites ${cluster_name}
+
+  git-crypt unlock
+  get_sudo
+
+  create_cluster ${cluster_name}
+  run_kops ${cluster_name}
+  install_components ${cluster_name}
+
+  kubectl cluster-info
+}
+
+create_cluster() {
+  local readonly cluster_name=$1
+  cd terraform/cloud-platform
+  switch_terraform_workspace ${cluster_name}
+  terraform apply -auto-approve
+}
+
+run_kops() {
+  local readonly cluster_name=$1
+
+  kops create -f kops/${cluster_name}.yaml
+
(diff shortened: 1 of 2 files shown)


create-cluster.sh
```diff
@@ -0,0 +1,179 @@
...
+  )
+}
+
+run_kops() {
+  local readonly cluster_name=$1
+
+  kops create -f kops/${cluster_name}.yaml
+
+  # This is a throwaway SSH key which we never need again.
+  rm -rf /tmp/${cluster_name}*
```
alkar: May I suggest simply `rm -f /tmp/${cluster_name}` to avoid any happy accidents? :)

example.env.create-cluster
```diff
@@ -0,0 +1,8 @@
+export AWS_PROFILE=moj-cp
+
+export KOPS_STATE_STORE=s3://cloud-platform-kops-state
```
alkar: `KOPS_STATE_STORE` is an output of the cluster creation terraform config (`cloud-platform/`). We should be able to compute it: `KOPS_STATE_STORE="s3://$(terraform output kops_state_store)"`. It shouldn't change unless we decide to move the state elsewhere but it would remove the need to have it exported beforehand. Do you think it's worth changing?

example.env.create-cluster
```diff
@@ -0,0 +1,8 @@
+export AWS_PROFILE=moj-cp
```
alkar: Is this strictly required? For the cluster creation bits we should have the appropriate profiles specified in the terraform provider configuration blocks.

create-cluster.sh
```diff
@@ -0,0 +1,179 @@
...
+check_env_vars() {
+  test $AWS_PROFILE
+  test $AUTH0_DOMAIN
+  test $AUTH0_CLIENT_ID
+  test $AUTH0_CLIENT_SECRET
+  test $KOPS_STATE_STORE
+}
+
+check_software_installed() {
+  hash git-crypt
```
alkar: Curious, why `hash` instead of `which`?

create-cluster.sh
digitalronin: There will be a key and a key.pub file. I'll change it to specify those explicitly.

create-cluster.sh
digitalronin: exit status

https://stackoverflow.com/a/677212/794111

example.env.create-cluster
digitalronin: I think we need it for kops anyway - I can't see any way to supply the aws profile to the kops command. If we need it for that, we may as well use it for terraform too.

example.env.create-cluster
digitalronin: The previous cluster build process computed it. The problem is that I don't think it's possible to export an env. var from *within* a script, and have it available for sub-shells of that script. So, I specified that it should be in the environment.

create-cluster.sh
alkar: Interesting, thanks for the link that's a treasure of an answer. 
FWIW, my `which` returns an exit code just fine and `hash` does too on `zsh`.

create-cluster.sh
alkar: (I was more concerned about the `-r` flag, not the wildcard)

example.env.create-cluster
alkar: You should be able to `export` from the parent shell but not the other way around. In any case, I don't think the value will change anytime soon so we can leave it as-is.

create-cluster.sh
digitalronin: Yeah, that was a mistake - just muscle memory!

create-cluster.sh
digitalronin: I think it's one of those things that will almost always work, but the answer also states that hash is a built-in, and which spawns a separate process, so it's arguably more efficient, and theoretically more reliable. There's very little in it. I think I'll add that link as a comment, to pre-empt future confusion!

PR #294, opened 2019-06-21
Add Cloudwatch Prometheus Exporter

jasonBirchall:
**Overview**
---
This PR links to https://github.com/ministryofjustice/cloud-platform/issues/823 and relates to the additional Prometheus scrape of Cloudwatch resources. When the commits in this PR have been merged and deployed to live a Cloudwatch exporter pod will be created in the `monitoring` namespace.

**How is this implemented**
---
Using the stable Helm chart, the [Prometheus-cloudwatch-exporter](https://hub.helm.sh/charts/stable/prometheus-cloudwatch-exporter) takes a set of metrics defined in a configmap and (using AWS credentials) scrapes data for Prometheus. The exporter requires readonly AWS credentials so we used KIAM to assume a role for relevant lookups. 

**Why**
---
There is a requirement to create triggered alerts on resources created outside the cluster. This particular PR for example, will allow developers to create `prometheusRules` that allow them to send a slack notification when latency exceeds x.

**What's next**
---
Just merging this PR into master won't actually deploy the Cloudwatch exporter to live-1. We need to perform a `terraform apply` on the `cloud-platform-components` directory. We also require documentation, both for the users and for cloud-platform engineers. These will come in separate PR's.


terraform/cloud-platform-components/resources/cloudwatch-exporter.yaml
```diff
@@ -0,0 +1,81 @@
...
+aws:
+  # The name of a pre-created secret in which AWS credentials are stored. When
+  # set, aws_access_key_id is assumed to be in a field called access_key,
+  # aws_secret_access_key is assumed to be in a field called secret_key, and the
+  # session token, if it exists, is assumed to be in a field called
+  # security_token
+  # Secret deployed from monitoring namespace in environments
+  secret:
+    name: "cloudwatch-prometheus-secret"
+    includesSessionToken: false
```
jasonBirchall: This secret is created in the environments repository. PR to follow. There is a question of whether this should exists as a Terraform resource?

terraform/cloud-platform-components/resources/cloudwatch-exporter.yaml
```diff
@@ -0,0 +1,81 @@
...
+  
+  - aws_namespace: AWS/RDS
+    aws_metric_name: WriteIOPS
+    aws_dimensions: [DBInstanceIdentifier]
+    aws_statistics: [Average]
+  
+  - aws_namespace: AWS/RDS
+    aws_metric_name: WriteLatency
+    aws_dimensions: [DBInstanceIdentifier]
+    aws_statistics: [Average]
```
jasonBirchall: I've only added RDS metrics for now, as this would be a good starting point.

terraform/cloud-platform-components/templates/prometheus-operator.yaml.tpl
```diff
@@ -1190,7 +1190,15 @@ prometheus:
     ## The scrape configuraiton example below will find master nodes, provided they have the name .*mst.*, relabel the
     ## port to 2379 and allow etcd scraping provided it is running on all Kubernetes master nodes
     ##
-    additionalScrapeConfigs: []
+    additionalScrapeConfigs:
```
alkar: Since you've enabled the `ServiceMonitor` in the helm chart, we shouldn't need this extra config block here.

terraform/cloud-platform-components/resources/cloudwatch-exporter.yaml
alkar: Can we not provide the necessary credentials with a role, using `kiam`?

terraform/cloud-platform-components/prometheus.tf
```diff
@@ -152,3 +152,18 @@ resource "helm_release" "alertmanager_proxy" {
     ignore_changes = ["keyring"]
   }
 }
+
+resource "helm_release" "cloudwatch_exporter" {
+  name      = "cloudwatch-exporter"
```
alkar: Also, since this is only really useful for live clusters, should we add something like `count = "${terraform.workspace == local.live_workspace ? 1 : 0}"` ?
If not, all test clusters will hammer the cloudwatch API potentially causing throttling issues.

terraform/cloud-platform-components/templates/prometheus-operator.yaml.tpl
jasonBirchall: A very good point. Reverted changed and padded out service monitor creation in values file. 
3b099f3

terraform/cloud-platform-components/prometheus.tf
jasonBirchall: Think this is a really good idea and is now implemented under commit 289a17e.

jasonBirchall:
@alkar thanks for the comments. Made requested changes, can you review once more?

terraform/cloud-platform-components/resources/cloudwatch-exporter.yaml
jasonBirchall: Yes, changed to reflect the above comment as this should assume a role via KIAM.

PR #295, opened 2019-06-21
3-AZ audit ES cluster & various fixes

alkar:
Connects to https://github.com/ministryofjustice/cloud-platform/issues/979


terraform/global-resources/elasticsearch.tf
```diff
@@ -324,4 +337,10 @@ resource "aws_elasticsearch_domain" "test" {
   ]
 }
 TESTPOLICY
+
+  log_publishing_options {
```
digitalronin: Isn't this a duplicate of lines 250-253?

alkar: It's on a different es cluster: `audit_1` vs `test`

terraform/global-resources/auth0.tf
```diff
@@ -5,13 +5,6 @@ resource "auth0_rule" "whitelist-github-orgs" {
   enabled = true
 }
 
-resource "auth0_rule" "whitelist-github-teams" {
-  name    = "whitelist-github-teams"
-  script  = "${file("${path.module}/resources/auth0-rules/whitelist-github-teams.js")}"
-  order   = 20
-  enabled = true
-}
```
jasonBirchall: What was the thinking here?

jasonBirchall: Do we no longer need to whitelist the team, or is this done via the org script now?

alkar: We do not use this rule, I think it was in place when everything was locked down to our team. It's not defined on the tenant.

jasonBirchall: :+1: okay, thanks.

PR #303, opened 2019-06-27
Template Slack receiver configuration for Alertmanager

alkar:
Our current method does not scale well, is very manual and error-prone. Instead, we can template this part of the config.

`terraform plan` against `live-1` show minimal changes that would not alter behaviour.


sablumiah:
all good expect line 28 of tfvars has an erroneous value

alkar:
@sablumiah thanks, it should be fixed now!

PR #310, opened 2019-07-04
replace test ES cluster enpoint with placeholder string

mtrblmoj:
- Replaced the test-elasticsearch endpoint that's applied to all non-live cluster 
- The new endpoint is a placeholder string, which seems to stop Fluentd from crashing
- If the test cluster has to be used, that string would have to be replaced manually.


mtrblmoj:
Next Step : having a working ES cluster in a pod, and being able to enable/disable its installation

PR #318, opened 2019-07-10
Ruby build script

digitalronin:
Replace the bash cluster build script with a ruby version.


create-cluster.rb
```diff
@@ -0,0 +1,188 @@
...
+#
+require "fileutils"
+
+MAX_CLUSTER_NAME_LENGTH = 12
+CLUSTER_SUFFIX = "cloud-platform.service.justice.gov.uk"
+# TODO: use the right dns flush command, depending on the architecture of the local machine
+DNS_FLUSH_COMMAND = 'sudo killall -HUP mDNSResponder' # Mac OSX Mojave
+
+REQUIRED_ENV_VARS = %w( AWS_PROFILE AUTH0_DOMAIN AUTH0_CLIENT_ID AUTH0_CLIENT_SECRET KOPS_STATE_STORE )
+REQUIRED_EXECUTABLES = %w( git-crypt terraform helm aws kops ssh-keygen )
```
jasonBirchall: This doesn't work for me, I get `./create-cluster.rb:131:in `block in check_software_installed': ERROR Required executable git-crypt not found. (RuntimeError)`

jasonBirchall:
```
$ which git-crypt
/usr/bin/git-crypt
```

jasonBirchall: How does this check required executables?

digitalronin: Interesting. It does the same thing as the bash version. Lines 130 & 131 iterate through a list of required executables, and then run 'hash [whatever]' which is the same thing as the bash script does.

What do you get for `hash git-crypt` ?

jasonBirchall: <chat continued in another medium>

PR #327, opened 2019-07-17
Nginx default error

vijay-veeranki-moj:
fixes https://github.com/ministryofjustice/cloud-platform/issues/1046


terraform/cloud-platform-components/nginx-ingress-acme.tf
```diff
@@ -180,3 +182,37 @@ resource "null_resource" "nginx_ingress_servicemonitor" {
     contents = "${sha1(file("${path.module}/resources/nginx-ingress/servicemonitor.yaml"))}"
   }
 }
+
+resource "null_resource" "nginx_ingress_errors_deployment" {
+  depends_on = ["helm_release.prometheus_operator"]
```
jasonBirchall: I might be missing something but why does it depend on prometheus?

vijay-veeranki-moj: Thanks @jasonBirchall , fixed it to depends on  to "helm_release.nginx_ingress_acme"

solidgoldpig:
Ouch. This clobbers any custom 400, 403 and 404 pages that an app might return. (502 and 504 almost by definition won't be returned by an app)

There needs to be some check to determine whether the app did send content along with the status code.

And if that's impossible to determine, 400, 403 and 404 pages should be not be attempted.

PR #341, opened 2019-08-12
Add a test that all namespaces are annotated

digitalronin:
This test currently fails when run against live-1, but that's because
namespaces should be annotated, but aren't. So, the test is correct, but
live-1 is wrong.


digitalronin:
The git history has become a bit messy, so I'm closing this PR in favour of #350

PR #345, opened 2019-08-15
Reduce name length for cluster backup checker role

digitalronin:
Previously, the name variable for this role evaluated to [cluster name
length] + 54 characters

This only leaves 10 characters for the cluster name. The build script
enforces a limit of 12 characters, so this commit reduces the name of the
iam role by 2 characters, so that no additional restrictions are added.


digitalronin:
Oops. Looks like I had an out of date working copy. I'll create a new PR.

PR #364, opened 2019-08-29
WIP Route53 helper

mtrblmoj:
Helper methods to support RSpec tests:

- create_zone
- create_delegation_set
- get_zone_records

Still missing: 
- delete_zone

### DIFF
--- a/smoke-tests/Gemfile
+++ b/smoke-tests/Gemfile
@@ -8,4 +8,4 @@ gem "rspec"
 gem "pry-byebug"
 gem "aws-sdk-core", "~> 3.59"
 gem "aws-sdk-iam", '~> 1.3'
-gem "aws-sdk-route53", '~> 1.27.0'
+gem "aws-sdk-route53", '~> 1.27'
--- a/smoke-tests/Gemfile
+++ b/smoke-tests/Gemfile
@@ -8,3 +8,4 @@ gem "rspec"
 gem "pry-byebug"
 gem "aws-sdk-core", "~> 3.59"
 gem "aws-sdk-iam", '~> 1.3'
+gem "aws-sdk-route53", '~> 1.27.0'
--- a/smoke-tests/spec/route53_helper.rb
+++ b/smoke-tests/spec/route53_helper.rb
@@ -1,0 +1,54 @@
+# Create new Route53 zone
+# Expect a domain name in input
+# Return route53 zone object https://docs.aws.amazon.com/sdkforruby/api/Aws/Route53/Types/CreateHostedZoneResponse.html
+def create_zone(domain)
+  client = Aws::Route53::Client.new
+  client.create_hosted_zone(
+    name: domain, # required
+    caller_reference: readable_timestamp, # required, different each time
+    hosted_zone_config: {
+      comment: "FOR TESTING PURPOSES ONLY",
+      private_zone: false,
+    },
+  )
+end
+
+# Delegate a Route53 zone (child -> Parent)
+# Expect a Route53 zone object and the parent zone_id
+def create_delegation_set(child_zone, parent_id)
+  client = Aws::Route53::Client.new
+  client.change_resource_record_sets(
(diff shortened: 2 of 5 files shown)


smoke-tests/spec/route53_helper.rb
```diff
@@ -0,0 +1,61 @@
+# Create new Route53 zone
+# Expect a domain name in input
+# Return route53 zone object https://docs.aws.amazon.com/sdkforruby/api/Aws/Route53/Types/CreateHostedZoneResponse.html 
+def create_zone(domain)
+    client = Aws::Route53::Client.new()
+    new_zone = client.create_hosted_zone({
```
digitalronin: You don't need to assign to a variable and then put the variable at the end of the function, i.e.
```
def foo
  var = do_something
  var
end
```
...is equivalent to
```
def foo
  do_something
end
```

Also, you don't need the outermost `{}` here.
`create_hosted_zone(foo: bar)` is the same as `create_hosted_zone({foo: bar})`

smoke-tests/spec/route53_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    new_zone
+end
+
+# Delegate a Route53 zone (child -> Parent)
+# Expect a Route53 zone object and the parent zone_id
+def create_delegation_set(child_zone, parent_id)
+
+    client = Aws::Route53::Client.new()
+    resp = client.change_resource_record_sets({
```
digitalronin: You don't need `resp =` here, because you're not doing anything with `resp`

smoke-tests/spec/route53_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+
+end
+
+# Retrieves a list of records from an existing Route53 zones
+# Expect a zone_id in input
+# Returns an array of hashes {type, name, value} of records.
+def get_zone_records(zone_id)
+  
+    client = Aws::Route53::Client.new()
+    records = client.list_resource_record_sets({
```
digitalronin: You can drop the `{}` here.

smoke-tests/Gemfile
```diff
@@ -8,3 +8,4 @@ gem "rspec"
 gem "pry-byebug"
 gem "aws-sdk-core", "~> 3.59"
 gem "aws-sdk-iam", '~> 1.3'
+gem "aws-sdk-route53", '~> 1.27.0'
```
digitalronin: Do you specifically need `1.27.0`? If you change this to `1.27` you'll get point releases whenever you run `bundle upgrade`

smoke-tests/spec/route53_helper.rb
```diff
@@ -0,0 +1,54 @@
+# Create new Route53 zone
+# Expect a domain name in input
+# Return route53 zone object https://docs.aws.amazon.com/sdkforruby/api/Aws/Route53/Types/CreateHostedZoneResponse.html
+def create_zone(domain)
+  client = Aws::Route53::Client.new
+  client.create_hosted_zone(
+    name: domain, # required
```
digitalronin: I don't think you need these `# required` comments. 

If you try to call `create_zone` without supplying a parameter, you'll get a runtime error

smoke-tests/spec/route53_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+# Expect a zone_id in input
+# Returns an array of hashes {type, name, value} of records.
+def get_zone_records(zone_id)
+  client = Aws::Route53::Client.new
+  records = client.list_resource_record_sets(
+    hosted_zone_id: zone_id, # required
+  )
+
+  records.resource_record_sets.collect { |r| {type: r.type, name: r.name, value: r.resource_records.map { |item| item.value }} }
+  # example:
```
digitalronin: I think it would be better to move this example comment above the 'def' with the description of the method

PR #371, opened 2019-09-02
add delete_zone and delete_delegation_set

mtrblmoj:
- add deletion methods to route53_helper.rb
- delete_zone deletes a route53 zone
- delete_delegation_set deletes the set of NS record pointing from a parent to a child zone

### DIFF
--- a/smoke-tests/spec/route53_helper.rb
+++ b/smoke-tests/spec/route53_helper.rb
@@ -60,7 +60,7 @@ def delete_zone(zone_id)
   client = Aws::Route53::Client.new
   client.delete_hosted_zone(
     id: zone_id,
-  })
+  )
 end
 
 # Deletes a Delegation set (NS)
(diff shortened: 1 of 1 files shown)


smoke-tests/spec/route53_helper.rb
```diff
@@ -52,3 +52,34 @@ def get_zone_records(zone_id)
 
   records.resource_record_sets.collect { |r| {type: r.type, name: r.name, value: r.resource_records.map { |item| item.value }} }
 end
+
+# Deletes a hosted zone.
+# Expects a zone_id in input
+# Be careful
+def delete_zone(zone_id)
+  client = Aws::Route53::Client.new
+  client.delete_hosted_zone({
```
digitalronin: You don't need the inner {} here.

smoke-tests/spec/route53_helper.rb
```diff
@@ -52,3 +52,34 @@ def get_zone_records(zone_id)
...
+          action: "DELETE",
+          resource_record_set: {
+            name: child_zone.hosted_zone.name,
+            resource_records: child_zone.delegation_set.name_servers.map { |ns| {value: ns} },
+            ttl: 60,
+            type: "NS",
+          },
+        },
+      ],
+      comment: "FOR TESTING PURPOSES ONLY",
```
digitalronin: I'd probably change this comment to "integrationtest", for the sake of having a single, consistent, uncommon string to search for.

PR #380, opened 2019-09-11
External DNS and R53 rspec

mtrblmoj:
This integration test introduces: 

- External DNS spec, testing External-DNS
- External DNS helper, containing key methods.
- Ingress.yaml template
- Route53 helper deletion methods


razvan-moj:
Can the zone IDs (ZD4.. ZQV..) and the ones used to test (child/parent) be moved to a single place with a 1-line explanation on why they are defined?

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  context "when zone matches ingress domain" do
+    let(:domain) { "child.parent.service.justice.gov.uk" }
+    let(:namespace) { "child-parent" }
+    let(:ingress_domain) { domain }
+    let(:parent_zone_id) { "ZQVC43X15AWL9" }
+    zone = nil
+
+    # Create a new zone per test
+    before do
+      zone = create_zone(domain)
```
digitalronin: Since you're not using `before(:all)` I think you might be able to do this with `let` statements. Have you tried that?

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        create_namespace(namespace)
+      end
+
+      after do
+        delete_namespace(namespace)
+      end
+
+      # an A record should be created
+      it "creates an A record" do
+        create_ingress
```
digitalronin: Since you mention the ingress in the context description, I'd probably put this in the before block, to keep the test easier to read

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      # When the ingress is deleted
+      context "when ingress is deleted" do
+        before do
+          delete_ingress(namespace) # includes waiting for confirmation
+        end
+
+        # The existing record in the zone should not deleted
+        it "does not delete records" do
+          records = get_zone_records(zone.hosted_zone.id)
+          expect(records).to_not be_nil
```
digitalronin: You probably need something like `records.count > 0` here, otherwise your test might pass when it shouldn't (e.g. `[].nil? == false`

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        end
+      end
+    end
+  end
+
+  # When no Route53 Zone match the ingress domain
+  context "when zone does not match ingress domain" do
+    let(:domain) { "otherchild.parent.service.justice.gov.uk" }
+    let(:namespace) { "child-parent" }
+    let(:parent_zone_id) { "ZQVC43X15AWL9" }
```
digitalronin: Is this a magic zone that must exist, in order for these tests to pass? Can we do something to create these resources only if they don't already exist, so that the tests always work, even if these records have been deleted?

smoke-tests/spec/fixtures/external-dns-ingress.yaml.erb
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: ingress-external-dns
```
digitalronin: See earlier comment about hard-coded values.

smoke-tests/spec/external_dns_helper.rb
```diff
@@ -1,31 +1,32 @@
-def create_ingress
+def create_ingress(namespace, ingress_name, fixture_name)
+  sleep 1
```
digitalronin: For the "sleep 1" at the start of methods, please add a comment explaining why it's there (because it looks like it's unnecessary). I'd put something like:

> \# TODO: sleep added to avoid AWS Route53 API throttling errors. Remove once that issue is resolved.

smoke-tests/spec/external_dns_helper.rb
```diff
@@ -1,31 +1,32 @@
...
   client = Aws::Route53::Client.new
 
   a_record = {
     action: "DELETE",
     resource_record_set: {
-      name: zone_name,
+      name: domain_name,
       alias_target: {
         "hosted_zone_id": "ZD4D7Y8KGAS4G",
```
digitalronin: Where does `ZD4D7Y8KGAS4G` come from?

smoke-tests/spec/external_dns_helper.rb
```diff
@@ -40,16 +41,17 @@ def delete_A_record(zone_id, zone_name, domain_name)
   })
 end
 
-def delete_TXT_record(zone_id, zone_name)
+def delete_TXT_record(zone_id, zone_name, domain_name, namespace)
```
digitalronin: Please change the method name to `delete_txt_record`, to follow ruby convention on method names

smoke-tests/spec/external_dns_helper.rb
```diff
@@ -0,0 +1,80 @@
...
+  else
+    delete_A_record(zone.hosted_zone.id, zone.hosted_zone.name, a_record_name)
+    delete_TXT_record(zone.hosted_zone.id, zone.hosted_zone.name)
+    delete_zone(zone.hosted_zone.id)
+  end
+end
+
+def is_zone_empty?(zone_id)
+  records = get_zone_records(zone_id)
+  !(records.size > 2)
```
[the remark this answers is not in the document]
digitalronin: ☝️

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -1,39 +1,44 @@
 require "spec_helper"
 
 describe "external DNS" do
-  # let!(zone) { nil }
+  namespace = "child-#{readable_timestamp}"
```
digitalronin: Can we have "integrationtest" in the namespace name? That matches the other tests, and will make it easy for us to identify and cleanup any leftovers (e.g. from crashed test runs)

smoke-tests/spec/external_dns_spec.rb
digitalronin: ☝️

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -1,39 +1,44 @@
...
 
-      after do
+      after(:all) do
         delete_namespace(namespace)
       end
 
       # an A record should be created
       it "creates an A record" do
-        create_ingress
+        create_ingress(namespace, ingress_name, fixture_name)
```
digitalronin: Given that the context name mentions that an ingress is created, I would move the `create_ingress` call, and the associated sleep, into the `before(:all)` block

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -1,39 +1,44 @@
 require "spec_helper"
 
 describe "external DNS" do
-  # let!(zone) { nil }
+  namespace = "child-#{readable_timestamp}"
+  let(:domain) { "child.parent.service.justice.gov.uk" }
+  let(:ingress_domain) { domain }
```
digitalronin: I think these statements belong in the "when zone matches ingress domain" context.

Ideally, the thing that makes the context different from other contexts should be done at the top of that context, to make it crystal clear to the reader what you're testing.

smoke-tests/spec/external_dns_helper.rb
mtrblmoj: It seems to be the zone where all the ELB domains are defined in eu-west-2, by AWS.
https://docs.aws.amazon.com/general/latest/gr/rande.html

PR #381, opened 2019-09-11
Increase fluentd container memory resource limit

digitalronin:
Fluentd containers on the master nodes are getting
OOM killed. This change increases the allowed
memory from 500Mi to 1000Mi to attempt to fix this


digitalronin:
This change has been applied by running terraform apply in the components folder.

PR #387, opened 2019-09-12
Adjust resource request/limit for fluentd pods

digitalronin:
This change had no effect:
https://github.com/ministryofjustice/cloud-platform-environments/commit/b79fb5505fd3689613c813752aafdd2d23c11fcb#diff-16341c618c10e25b6523a3767819dc66

This is because values in the helm chart specify
resources for the fluentd containers, overriding
the namespace default values in the limitrange.

This change alters the resource requests and
limits for the logging containers to use values
which are more representative of how they run in
production.

After this change, the daemonset will need to be
redeployed using a targeted terraform taint &
apply


digitalronin:
This PR is fine to merge. I was getting confused about a different change I made before this.

PR #392, opened 2019-09-13
Change fluentd containers resource requests

digitalronin:
* reduce CPU request from 100m to 10m
* increase memory request from 200Mi to 300Mi

This is closer to the resources the pods use on live-1, and will
enable us to reduce the resourcequota CPU request limit for the
namespace.


digitalronin:
This is a dupe of #387

PR #393, opened 2019-09-13
Added integration tests for default-backend

vijay-veeranki-moj:
This is related to https://github.com/ministryofjustice/cloud-platform/issues/1151

### DIFF
--- a/smoke-tests/spec/error_page_spec.rb
+++ b/smoke-tests/spec/error_page_spec.rb
@@ -1,9 +1,9 @@
 require "spec_helper"
 
-describe "custom error pages" do
+describe "cluster default 404 page" do
   context "404" do
     let(:url) do
-      "https://#{random_string}.apps.#{current_cluster}"
+      "https://foobar.apps.#{current_cluster}"
     end
 
     it "gets a 404 status" do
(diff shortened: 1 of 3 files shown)


smoke-tests/spec/error_page_spec.rb
```diff
@@ -12,11 +12,142 @@
       }.to raise_error(OpenURI::HTTPError, "404 Not Found")
     end
 
-    xit "serves a 404 response" do
-      URI.open(url)
-    rescue OpenURI::HTTPError => e
-      body = e.io.string
-      expect(body).to eq("404 page not found\n")
+    it "serves a 404 response" do
+      begin
```
digitalronin: I'd do this the same way as on lines 51 - 59, just for the sake of keeping things consistent.

smoke-tests/spec/error_page_spec.rb
```diff
@@ -12,11 +12,142 @@
...
+        body = e.io.string
+        expect(body).to include("Error, service unavailable - GOV.UK")
+      end
     end
   end
 end
+
+
+describe "Namespace with no custom default-backend and ingress annotations, serve cluster default-backend error page" do
+  let(:namespace) { "def-backend-#{readable_timestamp}" }
```
digitalronin: Objects created by the tests should be named "integrationtest-*"

This makes it a lot easier to find and delete them, if anything gets left behind (e.g. because a test run crashes)

smoke-tests/spec/kubernetes_helper.rb
```diff
@@ -169,3 +169,14 @@ def get_nodes
 def get_daemonsets
   JSON.parse(`kubectl get daemonsets --all-namespaces -o json`).fetch("items")
 end
+
+# Scale replicas
+def scale_replicas(namespace, deployment, replicas = "")
```
digitalronin: `replicas = ""` implies that you should supply a string value, but it makes more sense for this to be an integer.

As someone calling this method, I think it's reasonable to always require a value here, so I'd remove the default altogether.

smoke-tests/spec/kubernetes_helper.rb
```diff
@@ -169,3 +169,14 @@ def get_nodes
...
+  `kubectl -n #{namespace} scale deployment #{deployment} --replicas=#{replicas}`
+end
+
+# ingress_annotations
```
digitalronin: This comment doesn't add anything. I'd remove it.

smoke-tests/spec/kubernetes_helper.rb
```diff
@@ -169,3 +169,14 @@ def get_nodes
 def get_daemonsets
   JSON.parse(`kubectl get daemonsets --all-namespaces -o json`).fetch("items")
 end
+
+# Scale replicas
```
digitalronin: This comment doesn't add anything. I'd remove it.

PR #406, opened 2019-09-23
Add disable alerts

poornima-moj:
What:
Added script to find pagerduty_config and slack_webhook and replace with dummy values

Why:
To disable alerts when building test clusters by default.

### DIFF
--- a/create-cluster.rb
+++ b/create-cluster.rb
@@ -77,11 +77,9 @@ def install_components(cluster_name)
 end
 
 def disable_alerts
-  # This will disable high-priority pagerduty alarms for your cluster by replacing the token XXXXX with a dummy value 
-  # in line pagerduty_config = "XXXXX" in your working terraform.tfvars file.
+  # This will disable high-priority pagerduty alarms for your cluster by replacing the pagerduty_config token XXXXX with a dummy value
   `sed -i 's/pagerduty_config = ".*"/pagerduty_config = "dummydummy"/g' terraform/cloud-platform-components/terraform.tfvars`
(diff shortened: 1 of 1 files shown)


create-cluster.rb
```diff
@@ -58,6 +58,9 @@ def install_components(cluster_name)
   dir = "terraform/cloud-platform-components"
   execute "cd #{dir}; rm -rf .terraform"
   switch_terraform_workspace(dir, cluster_name)
+  
+  # Comment disable_alerts() to enable high priority - Pagerduty alarms and lower prority Slack alerts for your cluster
```
digitalronin: I'd move this comment from here, and down in the method body, add a comment above each 'sed' invocation, explaining its effect.

Also, will these leave your working copy in a state where there are outstanding changes?

poornima-moj: Yes, this will change the current working copy with an outstanding change. We have to ignore the  tfvars changes when pushing to git or checkout again after building the test cluster.

create-cluster.rb
```diff
@@ -75,6 +76,15 @@ def install_components(cluster_name)
   end
 end
 
+def disable_alerts
+  # This will replace the value of pagerduty_config to a dummy value in your working terraform.tfvars file.
```
digitalronin: I'd suggest something like this;

> This will disable high-priority pagerduty alarms for your cluster, by replacing the XXXX with a dummy value

...and something similar for the slack webhook.

I don't think the "comment this to enable..." line is necessary, as long as it's clear what effect the code has.

PR #419, opened 2019-09-26
upgrade externals-dns to v0.5.17

sablumiah:
upgrade both chart and app version for external-dns

~ helm_release.external_dns
    values.0:   image:
                -  tag: v0.5.11
                +  tag: 0.5.17-debian-9-r0
                 sources:
                   - service
                   - ingress
                 provider: aws
                 aws:
    version:    "1.7.0" => "2.6.4"

This update required a taint and apply due to a known issue:

https://github.com/operator-framework/operator-lifecycle-manager/issues/952

### DIFF
--- a/terraform/cloud-platform-components/external-dns.tf
+++ b/terraform/cloud-platform-components/external-dns.tf
@@ -41,7 +41,7 @@ resource "helm_release" "external_dns" {
   name      = "external-dns"
   chart     = "stable/external-dns"
   namespace = "kube-system"
-  version   = "2.6.4"
+  version   = "12.6.4"
 
   values = [<<EOF
 image:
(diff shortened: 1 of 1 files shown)


terraform/cloud-platform-components/external-dns.tf
```diff
@@ -66,12 +66,6 @@ podAnnotations:
 EOF
   ]
 
-  depends_on = [
```
digitalronin: IIRC, these were added in order to get the components to install in a single pass. 

What's the reason for removing the "depends_on" here?

sablumiah: Good catch Took them out for testing. Added back in now

PR #427, opened 2019-10-03
Build 1.13

razvan-moj:
**Why**
live cluster is upgraded to 1.13, new tests will be deployed with the same release

### DIFF
--- a/create-cluster.rb
+++ b/create-cluster.rb
@@ -21,7 +21,7 @@ def main(cluster_name)
 
   check_prerequisites(cluster_name)
 
-  execute "git-crypt unlock"
+  execute "git-crypt unlock ~/.aws/gitcrypt-infra"
  
   create_cluster(cluster_name)
   run_kops(cluster_name)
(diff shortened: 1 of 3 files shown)


create-cluster.rb
```diff
@@ -21,7 +21,7 @@ def main(cluster_name)
 
   check_prerequisites(cluster_name)
 
-  execute "git-crypt unlock"
+  execute "git-crypt unlock ~/.aws/gitcrypt-infra"
```
digitalronin: Why specify the path to the keyfile here? That path wouldn't work for me, whereas calling unlock without specifying a path works fine.

razvan-moj: This needs to be detailed separately in the runbook: when executing in a cluster, copying our private GPG keychain to the pod doesn't sound like a good idea, so instead we must export and `kubectl cp` the symmetric key. Same for AWS login credentials, a separate set not our own would be best.

digitalronin: OK, but for now we're not running this in the cluster - we're just mounting our own gpg files into the local container when we run it.
So, can we remove this change and just have the upgrade stuff in this PR?
We can do the key thing separately.

PR #442, opened 2019-10-22
Recycle node

vijay-veeranki-moj:
This is related to https://github.com/ministryofjustice/cloud-platform/issues/1153

This recycle-node.rb script will replace the oldest worker node.
-> Drain the oldest node
-> Terminate the drained node.

### DIFF
--- a/recycle-node.rb
+++ b/recycle-node.rb
@@ -1,16 +1,14 @@
 #!/usr/bin/env ruby
 
-# See example.env.create-cluster for the environment variables
-# which must be set before running this script.
 #
 # Usage:
-#   Change the K8S_CLUSTER_NAME and kubectl use-context of #K8S_CLUSTER_NAME
+#   export AWS_PROFILE=moj-cp
+#   Change the K8S_CLUSTER_NAME and set kubectl config to use-context #K8S_CLUSTER_NAME
 #   ./create-cluster.rb
 #
 
 require 'json'
 require "yaml"
-require "pry-byebug"  # TODO remove this
 require "net/http"
 
 K8S_CLUSTER_NAME = "vij-ing-fire.cloud-platform.service.justice.gov.uk"
@@ -28,6 +26,7 @@ def main
   node = get_oldest_worker_node
   drain_node(node)
   sleep 30
+  node = get_latest_node_details(node) # node status should have changed, after being drained
   terminate_node(node)
   sleep 60
 
@@ -82,6 +81,7 @@ def worker_node?(node)
 end
 
 def get_worker_instance_group_size
+  return 3
   docs = []
   YAML.load_stream(get_kops_config) { |doc| docs << doc }
   worker_instance_group = docs.last
@@ -104,7 +104,7 @@ def get_oldest_worker_node
 end
 
 def drain_node(node)
-  name = node.dig("metadata", "name")
+  name = node_name(node)
 
   cmd = "kubectl --ignore-daemonsets --delete-local-data drain #{name}"
 
@@ -113,6 +113,10 @@ def drain_node(node)
   else
     raise "worker node #{name} failed to drain. Aborting."
   end
+end
+
+def node_name(node)
+  node.dig("metadata", "name")
 end
 
 def terminate_node(node)
@@ -127,9 +131,14 @@ def aws_instance_id(node)
   node.dig("spec", "providerID").split("/").last
 end
 
+def get_latest_node_details(node)
+  name = node_name(node)
+  JSON.parse(execute("kubectl get node #{name} -o json"))
+end
+
 def cordoned?(node)
   spec = node.dig("spec")
-  spec.has_key?("unschedulable") && spec.dig("unschedulable") == "true"
+  spec.fetch("unschedulable", false)
 end
 
 def log(msg)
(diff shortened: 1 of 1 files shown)


recycle-node.rb
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env ruby
+
+
+# Usage:
+#   export AWS_PROFILE=moj-cp
+#   Edit "K8S_CLUSTER_NAME" to specify the cluster to recycle old node. 
+#   Set kubectl config to use-context #K8S_CLUSTER_NAME
```
digitalronin: The script can execute the `kubectl config` command, to ensure the context matches the cluster name. This is safer than requiring people to read the comments.

Also,  please remove the reference to `create-cluster.rb`

recycle-node.rb
```diff
@@ -0,0 +1,165 @@
...
+#   ./create-cluster.rb
+#
+#  Note: To run this on a test cluster, update function `get_worker_instance_group_size` as shown below,
```
digitalronin: Please add some explanation of why this is necessary.

Something like, "The method looks for the instance group size in the kops manifest in our github repository, so it will always return the size of the live-1 cluster..."

recycle-node.rb
```diff
@@ -0,0 +1,165 @@
...
+require 'json'
+require "yaml"
+require "net/http"
+
+K8S_CLUSTER_NAME = "recycle-node.cloud-platform.service.justice.gov.uk"
+AWS_REGION = "eu-west-2"
+KOPS_CONFIG_URL = "https://raw.githubusercontent.com/ministryofjustice/cloud-platform-infrastructure/master/kops/live-1.yaml"
+
+def main
+  number_of_workers = count_worker_nodes
```
digitalronin: line 25 doesn't do anything useful, since the logic testing the number is in `correct_number_of_workers_running?`

I'd also replace `worker_instance_group_size` with a call to `get_ worker_instance_group_size` on line 29, so you can remove line 26, too

recycle-node.rb
```diff
@@ -0,0 +1,165 @@
...
+end
+
+def count_worker_nodes
+  get_worker_nodes
+    .find_all { |node| node_ready?(node) }
+    .count
+end
+
+def get_nodes
+  JSON.parse(`kubectl get nodes -o json`).dig("items")
```
digitalronin: We should use `execute` here, rather than backticks, for consistency.

recycle-node.rb
```diff
@@ -0,0 +1,165 @@
...
+def get_latest_node_details(node)
+  name = node_name(node)
+  JSON.parse(execute("kubectl get node #{name} -o json"))
+end
+
+def terminate_node(node)
+  if cordoned?(node)
+    execute "aws ec2 terminate-instances --instance-ids #{aws_instance_id(node)} --region #{AWS_REGION}"
+  else
+    raise "Older worker node failed to terminate. Aborting."
```
digitalronin: This error should say "node was not cordoned..."

recycle-node.rb
```diff
@@ -0,0 +1,165 @@
...
+  end
+end 
+
+def aws_instance_id(node)
+  node.dig("spec", "providerID").split("/").last
+end
+
+def cordoned?(node)
+  spec = node.dig("spec")
```
digitalronin: The variable `spec` isn't doing much, here, I'd combine these as `node.dig("spec").fetch("unschedulable", false)`

recycle-node.rb
```diff
@@ -0,0 +1,165 @@
...
+def log(msg)
+  puts [Time.now.strftime("%Y-%m-%d %H:%M:%S"), msg].join(" ")
+end
+
+def cmd_successful?(cmd)
+  log cmd
+  system cmd
+end
+
+def execute(cmd, can_fail: false)
```
digitalronin: We're not using the `can_fail` flag here, so you can remove that from this method, to simplify the code.

vijay-veeranki-moj:
Thanks @digitalronin for the comments.
Updated considering review comments, can you please review it again.

recycle-node.rb
```diff
@@ -39,6 +39,10 @@ def main
   wait_for_node_to_be_replaced
 end
 
+def config_cluster
+  execute "kubectl config use-context #{K8S_CLUSTER_NAME}"
```
digitalronin: Sorry, one more thought. If this fails (e.g. there is no such context defined in the user's .kube/config file), the script will continue executing, and could delete a node in the wrong cluster.

This should have a guard where, unless the command returns successfully, we raise an error and die.

recycle-node.rb
```diff
@@ -155,10 +158,9 @@ def cmd_successful?(cmd)
   system cmd
 end
 
-def execute(cmd, can_fail: false)
+def execute(cmd)
   log cmd
-  result = `#{cmd}` # TODO: Use open3
-  raise "Command: #{cmd} failed." unless (can_fail || $?.success?)
+  result = `#{cmd}`
```
digitalronin: you can just end this method with the ``#{cmd}`` line - you don't need to assign to `result` .

vijay-veeranki-moj:
Thanks @digitalronin amended the code, can you please review it.

PR #462, opened 2019-11-18
Add Elasticsearch metrics

poornima-moj:
What and why:
Adding Elasticsearch module metrics to cloudwatch exporter so it can be monitored using Prometheus.

---------------- DONOT MERGE ----------
- This works fine when only scraping metrics for Elasticsearch. 

- Facing performance issues when scraping all namespaces. Cloudwatch exporter returns NullPointer error when scraping bulk metrics. We already have 89 metrics and ES add to another 59. 

Couldnot debug much using the Cloudwatch console as it shows wrong info. Support ticket raised with AWS to check any limits per API request.


poornima-moj:
Creating new PR with limited metrics for all AWS resources

PR #465, opened 2019-11-19
EKS for Tools/Management Cluster

mogaal:
This PR creates a directory hierarchy where it's going to be the infrastructure provisioning for the EKS. It follows exactly the same pattern than `terraform/cloud-platform/` with the following minor changes:

- Bastion was moved to a module in order to make it more reusable. For more details see ministryofjustice/cloud-platform#1434
- `eks.tf`: this is a new file which holds the initial EKS definition, we are using the official EKS module.
- New folder `files/` which holds the initial kubeconfig files for EKS, all files within this folder are encrypted using git-crypt (configuration added inside `.gitattributes`


terraform/cloud-platform-eks/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+  description = "a list of EC2 availability zones"
+  default     = ["eu-west-2a", "eu-west-2b", "eu-west-2c"]
+}
+
+variable "cluster_node_count" {
+  description = "The number of worker node in the cluster"
+  default     = "21"
+}
+
+variable "master_node_machine_type" {
```
mtrblmoj: I don't think this master_node var is required for eks.

mogaal: You are right, removed now

terraform/cloud-platform-eks/eks.tf
```diff
@@ -0,0 +1,39 @@
...
+      asg_desired_capacity = 2
+      key_name             = local.cluster_base_domain_name
+    }
+  ]
+
+  map_users = [
+    {
+      userarn  = "arn:aws:iam::754256621582:user/AlejandroGarrido"
+      username = "AlejandroGarrido"
+      groups   = ["system:masters"]
```
digitalronin: Is there any way to avoid having named individuals here? Can we have something that iterates through the members of the ministryofjustice/webops team, or a set of IAM users we pull from the AWS cli?

mogaal: I suppose we will try to implement our kuberos (or kubehook, which is not deprecated) so everyone will be able to connect using auth0. Still, we will need a seed user to apply the initial set up, this seed user is the one which will be listed there.

If we are going to choose the path you mention (our IAM accounts instead of auth0) will be good to do exactly what you said. But still we're going to need a seed user though (can be a generic user, it doesn't need to be an individual or member of our team).

I guess we'll realize which path will be when we start to deploy the cluster components,

mogaal: I'll start to deploy all cluster components we have for the kops and see how it goes with the kuberos

terraform/cloud-platform-eks/eks.tf
```diff
@@ -0,0 +1,39 @@
...
+      asg_max_size         = 5
+      asg_min_size         = 2
+      asg_desired_capacity = 2
+      key_name             = local.cluster_base_domain_name
+    }
+  ]
+
+  map_users = [
+    {
+      userarn  = "arn:aws:iam::754256621582:user/AlejandroGarrido"
```
razvan-moj: we'll need a solution for this

terraform/cloud-platform-eks/main.tf
```diff
@@ -0,0 +1,193 @@
...
+  cidr                 = var.vpc_cidr
+  azs                  = var.availability_zones
+  private_subnets      = var.internal_subnets
+  public_subnets       = var.external_subnets
+  enable_nat_gateway   = true
+  enable_vpn_gateway   = false
+  enable_dns_hostnames = true
+
+  public_subnet_tags = {
+    SubnetType                                                = "Utility" # <--- Why this is called Utility =(
```
razvan-moj: no good reason, feel free to change

terraform/cloud-platform-eks/main.tf
```diff
@@ -0,0 +1,193 @@
...
+    ),
+    format(
+      "https://kibana-audit.%s/oauth2/callback",
+      local.services_base_domain,
+    ),
+    format(
+      "https://grafana.%s/login/generic_oauth",
+      local.services_base_domain,
+    ),
+    format(
```
razvan-moj: we don't use this anymore, delete

terraform/cloud-platform-eks/variables.tf
```diff
@@ -0,0 +1,33 @@
...
+
+variable "availability_zones" {
+  type        = list(string)
+  description = "a list of EC2 availability zones"
+  default     = ["eu-west-2a", "eu-west-2b", "eu-west-2c"]
+}
+
+variable "cluster_node_count" {
+  description = "The number of worker node in the cluster"
+  default     = "21"
```
razvan-moj: might want a note in the README to create smaller test cluster

terraform/cloud-platform-eks/eks.tf
mogaal: Do you have something in mind? I can see within the IAM users you've one called "kops", we can create a new user that can be used as a "seed user" to access the first time to the cluster and deploy all components. Let me know what do you think

PR #474, opened 2019-11-28
Bugfix: correct the command which disables alerts

digitalronin:
We want to replace the pagerduty token and all
slack webhooks in test clusters, so we run a `sed`
command on the terraform source to replace those
values with `dummydummy`

But, because the code is executed via ruby, we
need to escape every backslash by doubling it
(i.e. wherever you put `\` in bash, e.g. when
running the command a terminal, you need to put
`\\` in your ruby code.

I think this is why we've still been getting 
alerts from our test clusters.

### DIFF
--- a/create-cluster.rb
+++ b/create-cluster.rb
@@ -116,10 +116,10 @@ end
 
 def disable_alerts
   # This will disable high-priority pagerduty alarms for your cluster by replacing the pagerduty_config token with a dummy value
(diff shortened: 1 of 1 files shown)


digitalronin:
The CircleCI validate is failing because https://github.com/ministryofjustice/cloud-platform-infrastructure/pull/476 hasn't been merged yet.

PR #475, opened 2019-11-28
remove the circleci config

digitalronin:
We have the github code formatter action, which does everything
the circleci job does, and which copes with terraform 0.12
syntax. So the circleci config isn't useful anymore.


digitalronin:
We also have OPA tests in the circle config. I'll raise a new PR which only removes the terraform formatting and validation steps

PR #479, opened 2019-12-04
sample configs for knative

razvan-moj:
also requires https://github.com/ministryofjustice/cloud-platform-tools-image/pull/48


digitalronin:
Raz has left, and we finished looking into knative for the time being, so I'm closing this PR.

PR #489, opened 2019-12-16
Configure kops to use a docker registry cache

digitalronin:
This change adds a docker registry to the live-1 cluster:
https://github.com/ministryofjustice/cloud-platform-environments/pull/1749

This commit alters the kops config, and the template for new kops
clusters, to use a registry cache via the URL provided.

Once this change has been merged, a rolling update of the live-1
cluster will be required to get worker nodes to start using the
cache.

### DIFF
--- a/kops/live-1.yaml
+++ b/kops/live-1.yaml
@@ -8,6 +8,8 @@ metadata:
   creationTimestamp: null
   name: live-1.cloud-platform.service.justice.gov.uk
 spec:
+  registryMirrors:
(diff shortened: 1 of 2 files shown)


digitalronin:
We want to restrict the cache to only be accessible from the worker nodes - not from the whole internet. So, this PR needs to wait on some more testing.

digitalronin:
This is wrong (should be `spec.docker.registryMirrors`), and a subsequent PR will include the correct version of this change, along with others to deploy the cache. So I'm closing this one.

PR #495, opened 2019-12-16
Split networking and updated components

mogaal:
There is a requirement to support multiple clusters inside the same VPC, for that we need to split networking components from the cluster. From now on the VPC creation happens inside the `cloud-platform-infrastructure/terraform/cloud-platform-network`

It was also bumped the versions of the components to match the latest one.


poornima-moj:
https://docs.google.com/document/d/1Z2bDB73kJPIsLJMKUrkRJJwPyOC8qv_iyOQZ0aPE-yc/edit?usp=sharing

PR #504, opened 2019-12-23
Add Velero Backup to cloud-platform-components

sablumiah:
WHAT
Add Velero backup to cloud-platform-components to install as parr of new cluster creation

WHY
Velero backup gives you tools to back up and restore your Kubernetes cluster resources and persistent volumes

### DIFF
--- a/terraform/cloud-platform-components/velero.tf
+++ b/terraform/cloud-platform-components/velero.tf
@@ -1,0 +1,138 @@
+# -----------------------------------------------------------
+# set up S3 bucket
+# -----------------------------------------------------------
+
+resource "aws_s3_bucket" "velero-test" {
+  #  provider      = "aws.cloud-platform-ireland"
+  #  bucket_prefix = "${var.bucket_prefix}"
+  acl    = "private"
+  region = "eu-west-2"
+
+  lifecycle {
+    prevent_destroy = false
+  }
+
+  server_side_encryption_configuration {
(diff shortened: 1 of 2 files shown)


terraform/cloud-platform-components/velero.tf
```diff
@@ -0,0 +1,100 @@
+resource "kubernetes_namespace" "velero" {
+  metadata {
+    name = "velero"
+
+    labels = {
+      "component" = "velero"
+    }
+
+    annotations = {
```
digitalronin: Should we add "is-production" "true" here?

terraform/cloud-platform-components/velero.tf
```diff
@@ -0,0 +1,100 @@
...
+    actions = [
+      "sts:AssumeRole",
+      "ec2:DescribeSnapshots",
+      "ec2:DescribeVolumes",
+      "ec2:CreateTags",
+      "ec2:CreateVolume",
+      "ec2:CreateSnapshot",
+      "ec2:DeleteSnapshot",
+    ]
+    resources = ["*"]
```
digitalronin: Is there any way to restrict this further, so that velero only has permissions over its own snapshots?

terraform/cloud-platform-components/velero.tf
sablumiah: added

PR #522, opened 2020-01-02
Deploy a docker registry as a pull-through cache

digitalronin:
closes https://github.com/ministryofjustice/cloud-platform/issues/1443

This PR deploys an instance of docker registry into the cluster, configured as
a pull-through cache (https://docs.docker.com/registry/recipes/mirror/).

To do this, we create a `docker-registry-cache` namespace, with a deployment
running a single replica of the registry. A health check is used to limit the
amount of disk space the cache consumes. Once the amount of disk space used in
the container exceeds a threshold, the health check will fail and the pod will
be terminated and replaced.

The worker nodes of the cluster are configured to use the registry mirror by
specifying a `registryMirrors` entry pointing to the ingress' public URL.  If
the registry is not available, the dockerd instance on the worker nodes will
simply fallback to fetching from docker hub directly, so there is no impact on
the cluster if the registry fails, and no dependency issue installing
components into the cluster before the registry is active.

The `nginx.ingress.kubernetes.io/whitelist-source-range` ingress annotation is
used to restrict access to the registry, so that it only serves requests which
originate from within the cluster VPC. To do this, we add the `nat_gateway_ips`
as an output when building the cluster VPC, and interpolate those IP numbers as
comma-separated CIDR ranges ("aa.bb.cc.dd/32") in the annotation.

NB: The split-components branch will need to be updated to output the IPs, and
to interpolate them correctly into the docker-registry-cache template.

Finally, this PR includes some integration tests to confirm that the cache is
running and correctly configured.

### Deploying

Once this PR is merged, we will need to do a `kops replace` and a `kops update`
as per [this runbook](https://runbooks.cloud-platform.service.justice.gov.uk/running-kops-update-rollingupdate.html)

It will not be necessary to do a `kops rolling-update`, since the node recycler
will eventually replace all the worker nodes with nodes configured to use the
cache, although we might want to accelerate this process to get the cluster
into a stable state as quickly as possible.

### DIFF
--- a/smoke-tests/spec/docker_registry_cache_spec.rb
+++ b/smoke-tests/spec/docker_registry_cache_spec.rb
@@ -1,0 +1,75 @@
+require "spec_helper"
+
+describe "docker-registry-cache" do
+  let(:namespace) { "docker-registry-cache" }
+
+  context "pod" do
+    let(:pods) { get_running_pods(namespace) }
+    let(:pod) { pods.first }
+
+    specify "one pod running" do
+      expect(pods.length).to eq(1)
+    end
+
+    specify "running the cache image" do
+      image = pod.dig("spec", "containers").first.fetch("image")
+      expect(image).to match("ministryofjustice.docker-registry-cache")
+    end
+
+    # Docker on the worker nodes should hit the docker-registry
+    # instance in the cluster whenever a container is launched.
+    # To test this, we check to see if any log output is generated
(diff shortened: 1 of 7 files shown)


digitalronin:
I'll merge and apply this tomorrow morning, so that I'm online in case anything goes awry.

digitalronin:
This has been merged, terraform applied, kops replaced, kops updated and kops rolling-updated.

PR #523, opened 2020-01-03
Commit changes to kops config after tf apply

digitalronin:
The latest terraform code adds
`spec.docker.registryMirrors` to the node config.

This commits those changes to the live-1 kops config file.


digitalronin:
This PR has additional changes in it, by mistake. Closing.

PR #532, opened 2020-01-07
Manager/Tool EKS cluster

mogaal:
Main change: Created the manager/tools cluster

It is also included in this change:
- Removed Raz from EKS users
- Removed test clusters mogaal and mogaal-01


mogaal:
Every `*.tfvars` file is encrypted by default. You can check it inside `.gitattributes` file, which is in the [root directory](https://github.com/ministryofjustice/cloud-platform-infrastructure/blob/master/.gitattributes).
Regarding the YAML file, the decision to encrypt everything inside `terraform/cloud-platform-eks/files/*` was taken because we didn't want to expose the `certificate-authority-data` (base64 cert use to connect to the K8S API endpoint), a good example is `terraform/cloud-platform-eks/files/kubeconfig_manager` inside the PR. I understand the fact that `config-map-aws-auth` doesn't include any secret or certs to connect to any endpoint but they are written together inside the same folder. Checking EKS module I can't find how to specify a custom path for the aws-map YAML file and the KUBECONFIG file.

PR #539, opened 2020-01-14
Change the ID of a namespace resource

digitalronin:
The docker-registry-cache namespace is defined in a kubernetes_namespace
resource which had the identifier "docker_registry_cache"

This was confusing the orphaned namespace checker, which reports that
the "docker-registry-cache" namespace is not defined in our source code.

Changing the identifier to match the namespace name should resolve this
problem (assuming terraform will allow an identifier with hyphens in
it).


digitalronin:
According to terraform plan, this will delete the whole namespace, and then recreate it with the same name. This shouldn't matter - it just means the cache will go away briefly, which it does anyway when the underlying disk space it uses grows too large.

PR #550, opened 2020-01-21
Add a 1-second delay to all calls, in pipeline

digitalronin:
If we're running in the concourse pipeline, add a 1-second delay before
every executed command, so  that we don't flood the kubernetes API and
cause KubeAPILatencyHigh alerts


digitalronin:
Hang on - It looks like I've got some extra commits that don't belong in here.

I'll fix it and raise another PR.

digitalronin:
Fixed by rebasing master. No need for another PR.

PR #574, opened 2020-02-11
Enable kubeScheduler and kubeControllerManager

vijay-veeranki-moj:
Updated with the right selectors for kubeScheduler and kubeControllerManager


vijay-veeranki-moj:
Thanks @digitalronin , updated the comment as suggested

PR #578, opened 2020-02-14
Thanos on live-1

mogaal:
In order to store Prometheus logs in S3 buckets, we need the Thanos sidecar. Now we have the latest prometheus-operator installed on live-1 we can deploy Thanos sidecar easily.

This PR has:

- _ThanosSpec_/configuration to the Prometheus Operator Values (`values.yaml`) as well as the IAM roles for the pod (to be able to write without credentials)
- IAM roles (and policies) creation for Thanos (through KIAM)
- The Kubernetes `side-car` and `thanos query` service resource
- The Kubernetes `thanos query` deployment

### DIFF
--- a/terraform/cloud-platform-components/prometheus.tf
+++ b/terraform/cloud-platform-components/prometheus.tf
@@ -261,6 +261,10 @@ resource "helm_release" "alertmanager_proxy" {
   }
 }
 
+##########
+# THANOS #
+##########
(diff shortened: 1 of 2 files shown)


mogaal:
This was moved out to its own module https://github.com/ministryofjustice/cloud-platform-terraform-prometheus

PR #585, opened 2020-02-18
Add tags, file per team and update role name to match application

poornima-moj:
- Contribution from Igor Ushkarev @ushkarev. 

- This will change the IAM role to more team specific by adding more tags, file name, and the name of the role.

- Add more permissions to the role to perform ACL based operations.


terraform/cross-account-IAM/iam-prisoner-money.tf
```diff
@@ -0,0 +1,61 @@
...
+
+variable "money-to-prisoners-test-tags" {
+  type = map(string)
+  default = {
+    business-unit    = "HMPPS"
+    application      = "money-to-prisoners"
+    is-production    = "false"
+    environment-name = "test"
+    owner            = "prisoner-money"
+  }
```
ushkarev: perhaps you also might want `infrastructure-support = var.infrastructure-support` as a line here

poornima-moj:
> Can you resubmit this as two separate commits, please? One commit to rename the file, and another commit to change it.
> 
> That will make it _much_ easier for whoever is reviewing it to see what the change is.

Sure @digitalronin. Will close and reopen with 2 PRs.

PR #589, opened 2020-02-24
Replace eks.tf

digitalronin:
related to #588

This commit replaces the eks.tf file. I'm hoping this will fix the
problem with git-crypt, described in #588


digitalronin:
The file hasn't been encrypted.

PR #590, opened 2020-02-24
Replace eks.tf (hopefully with encrypted version)

digitalronin:
related to #588


digitalronin:
still not encrypted :(

PR #591, opened 2020-02-24
Add eks.tf as eks-cluster.tf to fix git-crypt

digitalronin:
git-crypt remembers when a file you tell it to encrypt has existed
in the repo previously, in cleartext. So, commit e75f22ff had the
side-effect of breaking git-crypt and making it impossible to add
the eks.tf file encrypted.

This commit renames the file to eks-cluster, which has never
existed in the repo before, and so git-crypt should encrypt it as
normal.

related to #588


digitalronin:
still not encrypted - I don't know why

PR #599, opened 2020-02-26
Add tests for master nodes

digitalronin:
* 3 master nodes
* All running standard master node pods
* One running a dns-controller pod

### DIFF
--- a/smoke-tests/spec/master_nodes_spec.rb
+++ b/smoke-tests/spec/master_nodes_spec.rb
@@ -1,0 +1,70 @@
+require "spec_helper"
+
+describe "master nodes", speed: "fast" do
+
+  # normalise pod names for ease of comparison, e.g.
+  #
+  #   calico-node-mv48v -> calico-node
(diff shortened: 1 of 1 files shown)


smoke-tests/spec/master_nodes_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  # We should have 3 masters
+  specify do
+    expect(masters.size).to eq(3)
+  end
+
+  # Every master should have these pods running, in these namespaces
+  it "has standard pods" do
+    expected_pods = [
+      ["kube-system", "calico-node"],
```
jasonBirchall: You could also add calico-kube-controller.

jasonBirchall: Only appears as a single pod however.

digitalronin: There is only one of those, but it doesn't run on a master node (necessarily? it's running on a worker node right now - could it run on a master?)

PR #605, opened 2020-02-28
Extend abilities of cluster-creation script

mogaal:
In order to create the build-test-clusters pipeline, some extra functionalities `create-cluster.rb` script had to be added:

- Make the `git-crypt unlock` optional: When it is executed from the concourse pipeline the repository is already unlocked.
- Make integration tests optional: all of them are failing and for test clusters it looks it is expected
- Optional time (argument in seconds) between cluster validation (kops validate) and components deployment (`terraform apply` inside components folder). We were experimenting failing DNS resolutions and it looks it wasn't fully propagated the DNS API at the moment when components were deployed.

### DIFF
--- a/create-cluster.rb
+++ b/create-cluster.rb
@@ -44,6 +44,8 @@ def main(options)
   cluster_size = options[:cluster_size]
   vpc_name = options[:vpc_name]
   gitcrypt_unlock = options[:gitcrypt_unlock]
+  extra_wait = options[:extra_wait]
+  
 
   vpc_name = cluster_name if vpc_name.nil?
   usage if cluster_name.nil? || cluster_size.nil?
@@ -55,7 +57,7 @@ def main(options)
   create_vpc(vpc_name)
   create_cluster(cluster_name, cluster_size, vpc_name)
   run_kops(cluster_name)
-  sleep(600)
+  sleep(extra_wait) unless extra_wait.nil?
   install_components(cluster_name)
   run_integration_tests(cluster_name)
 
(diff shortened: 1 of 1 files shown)


create-cluster.rb
```diff
@@ -27,35 +27,39 @@
...
-  execute "git-crypt unlock"
+  execute "git-crypt unlock" if gitcrypt_unlock == true
 
   create_vpc(vpc_name)
   create_cluster(cluster_name, cluster_size, vpc_name)
   run_kops(cluster_name)
+  sleep(extra_wait) unless extra_wait.nil?
   install_components(cluster_name)
-  run_integration_tests(cluster_name)
+  run_integration_tests(cluster_name) if integration_tests == true
```
digitalronin: You could just say `if gitcrypt_unlock` and `if integration_tests` here - you don't need the ` == true`

`nil` is "falsy" in ruby, so you can also say `sleep(extra_wait) if extra_wait`

create-cluster.rb
```diff
@@ -267,6 +269,18 @@ def parse_options
       options[:vpc_name] = name
     end
 
+    opts.on("-g", "--no-gitcrypt", "Avoid the execution of git-crypt unlock (example: pipeline tools might do that for you)") do |name|
+      options[:gitcrypt_unlock] = false
+    end
+
+    opts.on("-i", "--no-integration-test", "Avoid the execution of git-crypt unlock (example: pipeline tools might do that for you)") do |name|
```
digitalronin: The comment here is incorrect, it should be something like "Don't run integration tests after creating the cluster"

create-cluster.rb
mogaal: Ahhh cool, thank you. Now improved. Please check

create-cluster.rb
mogaal: Yeah sorry, copy-pasted from the above one. Fixed now.

mogaal:
Ups, one sec. Some merge problems.

PR #620, opened 2020-03-09
Commented out ingress with invalid syntax test

vijay-veeranki-moj:
This is to verify if api latency high is caused because of this test.

Will un-do this change after validating the result, by running smoke tests.

### DIFF
--- a/smoke-tests/spec/ingress_spec.rb
+++ b/smoke-tests/spec/ingress_spec.rb
@@ -27,7 +27,7 @@ describe "nginx ingress" do
         namespace: namespace,
         host: host,
         file: "spec/fixtures/helloworld-deployment.yaml.erb",
-        binding: binding,
+        binding: binding
       )
 
       wait_for(namespace, "ingress", "integration-test-app-ing")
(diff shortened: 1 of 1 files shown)


smoke-tests/spec/ingress_spec.rb
```diff
@@ -40,15 +40,15 @@
...
-      stdout_str, stderr_str, status = apply_template_file(
-        namespace: namespace,
-        host: host,
-        file: "spec/fixtures/invalid-nginx-syntax.yaml.erb",
-        binding: binding,
-      )
-      expect(stderr_str).to match(/admission webhook "validate.nginx.ingress.kubernetes.io" denied the request/)
-    end
-  end
+  # context "when ingress is deployed with invalid syntax" do
```
digitalronin: @vijay-veeranki-moj you should be able to just change `context` to `xcontext` to achieve this.
That has the added benefit of reporting, in the test output, that the test is being skipped, so that we don't forget about it.

vijay-veeranki-moj: Thanks @digitalronin, made the change , can you please review it again

PR #632, opened 2020-03-19
Put kops worker nodes instancegroup at end of file

digitalronin:
The node recycler expects the main worker nodes instancegroup to be the last
document in the kops yaml file. This change puts it there, and adds an
explanatory comment.

This change includes both the template and the file which is produced by
`terraform apply`


digitalronin:
I've already done the `kops replace...` and `kops update cluster ...` with this version.

PR #634, opened 2020-03-19
Adding tolerations

mogaal:
- Updated Prometheus module version to one which supports tolerations.
- Updated Kiam with tolerations for big nodes (monitoring one)


mogaal:
This was already applied in another PRs.

PR #644, opened 2020-03-30
nginx upgrade

mogaal:
As part of the Helm 3 upgrade, we're facing the well-known [`spec.clusterIP` problem](https://github.com/helm/charts/pull/13646). Once we upgrade to the latest and start using `XXX.service.omitClusterIP=true` values, the error should disappear and we can carry on with the upgrade.

This PR sets all `omitClusterIP` to true and upgrade the Helm Chart to the latest one.


mogaal:
Nginx has been moved to its own terraform module, these changes are applied in [#656](https://github.com/ministryofjustice/cloud-platform-infrastructure/pull/656)

PR #684, opened 2020-04-28
Update elasticsearch.tf

vijay-veeranki-moj:
Increased the node size to "8", to fix "ClusterIndexWritesBlocked" issue due to storage space.


vijay-veeranki-moj:
@digitalronin yes it fixed the problem.

PR #694, opened 2020-05-01
Improve External-DNS integration test

mtrbls:
This PR tries to make the ext-dns test more resilient. 

Most issues with this test happens because of a badly cleaned up R53 zone. 
Before running any test, this spec will now remove any remaining records (A, TXT)  from the existing zone.


smoke-tests/spec/external_dns_helper.rb
```diff
@@ -1,3 +1,15 @@
+# Return the zone_id of a zone, found by name (domain)
+# This will only use the first result
+def get_zone_by_name(domain)
+  client = Aws::Route53::Client.new
+
+  zones = client.list_hosted_zones_by_name({
+    dns_name: domain,
+    max_items: 1
+  })
+  zone_id = zones.hosted_zones[0].id.tr("/hostedzone/", "")
```
digitalronin, line 10: You don't need the "zone_id = " here - whatever the last value evaluated in the function is, that's what the return value will be.

smoke-tests/spec/external_dns_helper.rb
```diff
@@ -78,21 +65,28 @@ def delete_txt_record(zone_id, domain_name, namespace)
...
+def cleanup_zone(domain, namespace, ingress_name, zone_id = nil)
+  if zone_id.nil?
+    zone_id = get_zone_by_name(domain)
+  end
+
+  records = get_zone_records(zone_id)
+
+  unless is_zone_empty?(records)
+    records.each do |record|
+      case record[:type]
```
digitalronin, line 77: Do you need different methods to delete TXT versus A records? If you're passing the record object into the delete function, can you replace "delete_a_record" and "delete_txt_record" with a single "delete_record" method?

smoke-tests/spec/external_dns_helper.rb
```diff
@@ -78,21 +65,28 @@ def delete_txt_record(zone_id, domain_name, namespace)
...
   end
 end
 
 # Checks if a zone is empty
 # A zone is considered empty if it only contains one SOA and one NS record
-# TODO: sleep added to avoid AWS Route53 API throttling errors. Remove once that issue is resolved.
-def is_zone_empty?(zone_id)
-  sleep 1
-  records = get_zone_records(zone_id)
+def is_zone_empty?(records)
```
digitalronin, line 89: If this method is only ever called from one place in the code, I would get rid of it and just inline `records > 2` as a boolean, with the explanatory comment next to it.

smoke-tests/spec/external_dns_spec.rb
```diff
@@ -3,14 +3,14 @@
 # This test can only be ran against live-1. Test clusters do not have enough privileges.
 describe "external DNS", "live-1": true do
   let(:domain) { "integrationtest.service.justice.gov.uk" } # That zone already exists
-  zone = "Z02429076QQMAO8KXV68" # integrationtest.service.justice.gov.uk zone_id
+  zone_id = "Z02429076QQMAO8KXV68" # integrationtest.service.justice.gov.uk zone_id
```
digitalronin, line 6: This should probably be defined as a constant, to clarify that it's a magic value.

smoke-tests/spec/external_dns_helper.rb
mtrbls: you are right, it is much simpler that way.

PR #703, opened 2020-05-06
Use the latest logging module

mtrbls:
In the new release of the logging module, the _fluentd_ pods are given more CPU resource. 

This should prevent the pods from running out of CPU, on the master nodes.


digitalronin:
@mtrbls can this be merged?

PR #724, opened 2020-05-27
removed unused vars. replaced assume policy with document policy reso…

imranawan-moj:
…urce


imranawan-moj:
this is now redundant as every team will have their ecr scan lambda function now as opposed to having it global.

PR #729, opened 2020-05-29
Fix JSON parse

mogaal:
The new ES API returns different JSON. Our integration tests need to be updated.


smoke-tests/spec/logging_spec.rb
```diff
@@ -39,7 +39,7 @@
     pod_name = pod.dig("metadata", "name")
     json = get_pod_logs(namespace, pod_name) # results from the elasticsearch query
     hash = JSON.parse(json)
-    total_hits = hash.fetch("hits").fetch("total")
+    total_hits = hash.fetch("hits").fetch("total").fetch("value")
```
digitalronin, line 42: You could do this like this:
```
total_hits = hash.dig("hits", "total", "value")
```

Although, that will behave differently if any of those keys are missing. `fetch` will raise an error, but the `dig` version will just return nil.

PR #758, opened 2020-06-17
Concourse tests

imranawan-moj:
New 'concourse_spec.rb' test added to verify concourse / concourse-main namespaces and running of pods within concourse ns.

### DIFF
--- a/smoke-tests/spec/concourse_spec.rb
+++ b/smoke-tests/spec/concourse_spec.rb
@@ -1,0 +1,35 @@
+require "spec_helper"
+
+describe "concourse-test", "eks-manager": true do
+  context "concourse namespace should exists" do
+    it "Fail if namespace doesnt exist" do
+      expect(namespace_exists?("concourse")).to eq(true)
+    end
+  end
+
+  context "expected all concourse-worker pods running inside concourse namespace" do
+    let(:pods) { get_running_app_pods("concourse", "concourse-worker") }
+
+    it "all concourse-worker pods inside concourse are running" do
+      expect(all_containers_running?(pods)).to eq(true)
+    end
+  end
+
+  context "expected all postgresql pods running inside concourse namespace" do
+    let(:pods) { get_running_app_pods("concourse", "postgresql") }
+
+    it "all postgresql pods inside concourse are running" do
+      expect(all_containers_running?(pods)).to eq(true)
+    end
+  end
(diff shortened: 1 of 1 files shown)


smoke-tests/spec/concourse_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+describe "concourse-test", "eks-manager": true do
+  context "concourse namespace should exists" do
+    it "Fail if namespace doesnt exist" do
+      expect(namespace_exists?("concourse")).to eq(true)
+    end
+  end
+
+  context "concourse-main namespace should exists" do
+    it "Fail if namespace doesnt exist" do
```
digitalronin, line 11: I'd probably remove the "contexts" here.

You usually add a context when you need to have different setups for different groups of tests. But, in this case, you're just testing for the existence of some namespaces, so the contexts don't add value.

Also, I'd probably change this
```
    it "Fail if namespace doesnt exist" do
      expect(namespace_exists?("concourse")).to eq(true)
```
to this:
```
  specify {
      expect(namespace_exists?("concourse")).to eq(true)
  }
```

`specify` and `it` are synonyms in rspec, and the code itself is simple enough that the descriptive text doesn't really make it any easier to read.

smoke-tests/spec/concourse_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  context "expected all concourse-web running inside concourse namespace" do
+    let(:pods) { get_running_app_pods("concourse", "concourse-web") }
+
+    it "all concourse-web pods inside concourse are running" do
+      expect(all_containers_running?(pods)).to eq(true)
+    end
+  end
```
digitalronin, line 38: Similarly, because you've only got one test per context here, I'd probably drop the contexts and change the tests to fetch the `pods` inside the tests, like this;
```
it "runs postgresql pods" do
  pods = get_running_app_pods("concourse", "postgresql")
  expect(all_containers_running?(pods)).to eq(true)
end
```

imranawan-moj: done.

smoke-tests/spec/concourse_spec.rb
imranawan-moj: done

PR #759, opened 2020-06-17
Wildcard Certificate in manager cluster

mogaal:
During the concourse migration as terraform module we realised the nginx default certificate in manager cluster does not include the wildcard certificate for `*.cloud-platform.service.justice.gov.uk`. So we had to add it manually in order to get a valid certificate for `concourse.cloud-platform.service.justice.gov.uk`.

This PR makes the change permanent.


mogaal:
Sorry, this line might cause problems with this https://github.com/ministryofjustice/cloud-platform-terraform-ingress-controller/blob/master/templates/values.yaml.tpl#L86

So both ExternalDNS might end-up fighting for the same. Further investigation

PR #760, opened 2020-06-19
Create pod security policy integration tests

jasonBirchall:
Overview
---
This PR connects to https://github.com/ministryofjustice/cloud-platform/issues/1908 and relates to the creation of integration tests for the pod security policies in a Cloud Platform cluster. The tests check the following:
- Whether the three important pod security policies exist in the cluster
- Whether a privileged and unprivileged container can run inside and privileged namespace.
- Whether a privileged container fails to run inside an unprivileged namespace.
- Whether an unprivileged container can run inside an unprivileged namespace.

### DIFF
--- a/smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
+++ b/smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
@@ -16,3 +16,4 @@ spec:
           imagePullPolicy: "IfNotPresent"
           ports:
             - containerPort: 5432
+
--- a/smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
+++ b/smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
@@ -1,10 +1,13 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: privileged-integration-test
   namespace: <%= namespace %>
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      app: privileged-integration-test
   template:
     metadata:
       labels:
--- a/smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
+++ b/smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
@@ -1,0 +1,18 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: privileged-integration-test
+  namespace: <%= namespace %>
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: privileged-integration-test
+    spec:
+      containers:
+        - name: postgres
+          image: postgres:10.4
+          imagePullPolicy: "IfNotPresent"
+          ports:
+            - containerPort: 5432
--- a/smoke-tests/spec/fixtures/unprivileged-deployment.yaml.erb
+++ b/smoke-tests/spec/fixtures/unprivileged-deployment.yaml.erb
@@ -16,3 +16,4 @@ spec:
           imagePullPolicy: "IfNotPresent"
           ports:
             - containerPort: 5432
+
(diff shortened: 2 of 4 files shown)


smoke-tests/spec/kubernetes_helper.rb
```diff
@@ -69,6 +69,24 @@ def object_exists?(namespace, type, name)
...
+end
+
+# Creates a deployment using the bitnami/nginx image with a unprivileged container.
+def create_unprivileged_deploy(namespace)
+  apply_template_file(
+    namespace: namespace,
+    file: "spec/fixtures/unprivileged-deployment.yaml.erb",
+    binding: binding
+  )
+end
```
digitalronin, line 88: These seem very specific to the PSP tests, so I'd probably put them inline in the psp_spec.rb file. But, since the only difference is the name of the ERB file, and the name includes "privileged" and "unprivileged", I'm not sure it's worth defining methods for these.

smoke-tests/spec/psp_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      create_namespace(namespace)
+      make_namespace_privileged(namespace)
+    end
+
+    after do
+      delete_namespace(namespace)
+      delete_clusterrolebinding(namespace)
+    end
+
+    it "privileged containers run" do
```
digitalronin, line 36: For the names of tests, generally you either want to use "it" to describe a situation, like 

> it "allows root containers to run" do ...

Or, if that produces names that are too clunky, you might use "specify" like this;

> specify "containers run as root" do ....

smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
```diff
@@ -0,0 +1,19 @@
+apiVersion: extensions/v1beta1
```
poornima-moj, line 1: @jasonBirchall , can we use apps/v1 as we are moving towards k8s 1.16

smoke-tests/spec/psp_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      "restricted",
+      "kube-system"
+    ]
+    expect(names).to include(*expected)
+  end
+
+  # Runs a privilege namespace and confirms both privilege and
+  # non-privilege containers can run. 
+  context "when namespace is privileged" do
+    before do
```
digitalronin, line 26: I'd probably use something in here to explain that the erb template is part of what distinguishes privileged and non-privileged namespaces, e.g.
```
context "when namespace is privileged" do
  let(:deployment_file) { "spec/fixtures/privileged-deployment.yaml.erb" }
  ...
```
And similarly for unprivileged.

That would mean reorganising things so that it's the *context* which is privileged or not, but I think that would make the tests clearer and a little more concise.

smoke-tests/spec/fixtures/unprivileged-deployment.yaml.erb
```diff
@@ -0,0 +1,19 @@
+apiVersion: extensions/v1beta1
```
poornima-moj, line 1: @jasonBirchall , same as above: replace with apps/v1

smoke-tests/spec/fixtures/privileged-deployment.yaml.erb
jasonBirchall: A very good point. I'll change both. :+1:

smoke-tests/spec/kubernetes_helper.rb
jasonBirchall: I agree with this, I've now moved it to the psp spec and made the function generic.
```
# Creates a deplyoment using the arguments defined
def create_psp_deployment(namespace, deployment_file)
  apply_template_file(
    namespace: namespace,
    file: deployment_file,
    binding: binding
  )
end

```

smoke-tests/spec/psp_spec.rb
jasonBirchall: Agreed. Changed the names of all descriptions.

smoke-tests/spec/psp_spec.rb
jasonBirchall: I originally used the organisation you outline but decided to change it to make as few function calls as possible. Reading your comment, I think you're right so I have decided to revert back.

PR #788, opened 2020-07-14
Updated kops module with kops 1.15.12 changes

mogaal:
In the previous PR to include the template changes done to kops manifest due to the upgrade against 1.15.12


mogaal:
Ohhh good point @jasonBirchall. I'm gonna start working on the docs. Thanks for reminding me, I would've totally missed it.

PR #813, opened 2020-07-23
AWS EKS Pod Security Policies

jasonBirchall:
This PR amends the Cloud Platform pod security policies, allowing them to work on an AWS EKS cluster (specifically the live manager cluster). It also includes the removal of the "kops only" flag in our integration tests, meaning it'll be run against both EKS 
and kops clusters in the future. 

The API groups had to be changed as the extensions API has been deprecated. In addition to this a number of service accounts were added to allow numerous kube-system pods access to the privileged PSP.

When merged we will need to perform a `terraform apply` to the components of both live-1 and manager. EKS comes with a default PSP, this will need to be removed in all future clusters. An issue has been raised to approach this work [here](https://github.com/ministryofjustice/cloud-platform/issues/2106)

This PR closes https://github.com/ministryofjustice/cloud-platform/issues/2066.

### DIFF
--- a/smoke-tests/spec/psp_spec.rb
+++ b/smoke-tests/spec/psp_spec.rb
@@ -14,8 +14,7 @@ describe "pod security policies", kops: true do
 
     expected = [
       "privileged",
-      "restricted",
-      "kube-system"
+      "restricted"
     ]
     expect(names).to include(*expected)
   end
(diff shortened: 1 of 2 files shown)


jasonBirchall:
Sorry @sablumiah I'd forgotten to push one more change, are you okay to re-review?

PR #839, opened 2020-08-11
Updated baselines to the up to date terraform module

mogaal:
Also, this PR clean up resources and provider configurations that are not in use anymore.

### DIFF
--- a/terraform/cloud-platform-account/main.tf
+++ b/terraform/cloud-platform-account/main.tf
@@ -9,22 +9,9 @@ terraform {
   }
 }
 
-locals {
-  workspace_to_profile = {
-    "cloud-platform"              = "moj-cp"
-    "mojdsd-platform-integration" = "moj-pi"
-  }
-}
-
 provider "aws" {
-  version = ">= 1.44.0"
   region  = "eu-west-2"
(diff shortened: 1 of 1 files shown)


terraform/cloud-platform-account/main.tf
```diff
@@ -9,22 +9,9 @@ terraform {
...
 provider "aws" {
-  version = ">= 1.44.0"
   region  = "eu-west-2"
-  profile = local.workspace_to_profile[terraform.workspace]
-}
-
-provider "aws" {
-  region = "eu-west-1"
-  alias  = "ireland"
+  profile = "moj-cp"
```
digitalronin, line 14: Do we have this requirement (i.e. that you must have an AWS profile called "moj-cp") documented anywhere?

digitalronin: Yes, we do. It's in the "create a cluster" runbook.

mogaal: I don't think so, but it is widely used across all our terraform files. For me they don't make much sense unless we use multiple AWS Accounts, but it isn't the case. 

With `AWS_*` exported should be more than enough

PR #860, opened 2020-08-19
Increase sleep delay to make tests more reliable

digitalronin:
This psp spec fails occasionally. Increasing sleep to 20 seconds seems
to have fixed the other psp tests, so hopefully this will have the same
effect for this one.


mtrbls (approved):

imran-awan1 (commented):
reviewed

PR #902, opened 2020-09-25
Define per-AZ instancegroups for worker nodes

vijay-veeranki:
This is related to [2272](https://github.com/ministryofjustice/cloud-platform/issues/2272)

This will help to alter our kops config to control which AZ our worker nodes are created.

Also updated the recycle node job script to handle multiple AZ ig.

### DIFF
--- a/recycle-node.rb
+++ b/recycle-node.rb
@@ -85,7 +85,7 @@ end
 
 def worker_node?(node)
   instance_group = node.dig("metadata", "labels", "kops.k8s.io/instancegroup")
-  
+
   node.dig("metadata", "labels")["kubernetes.io/role"] == "node" \
     && recylable_instance_groups.include?(instance_group)
 end
--- a/recycle-node.rb
+++ b/recycle-node.rb
@@ -22,7 +22,7 @@ AWS_REGION = "eu-west-2"
 KOPS_CONFIG_URL = "https://raw.githubusercontent.com/ministryofjustice/cloud-platform-infrastructure/main/kops/live-1.yaml"
 NODE_DRAIN_TIMEOUT = 360 # Draining a node usually takes around 2 minutes. If it takes >6 minutes, it's not going to complete.
 SIGTERM = 15 # The unix signal to send to kill a process
-WORKER_NODE_INSTANCEGROUP = "nodes-1.16.13" # The name of the worker nodes instancegroup in the kops config.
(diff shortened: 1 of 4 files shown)


digitalronin (changes requested):
Is there any way we can label the worker node instance groups with an additional label like "WORKER_NODES" or something?

It would be nice if we could come up with something where we don't need to keep editing the `recycle-node.rb` script every time we rename an instance group (e.g when we upgrade kubernetes).

If we had a "static" label in addition to the node group name, we could tell the script to pick a node to recycle from all the instancegroups which have that tag.

recycle-node.rb
```diff
@@ -80,19 +80,19 @@ def node_ready?(node)
...
   node.dig("metadata", "labels")["kubernetes.io/role"] == "node" \
-    && node.dig("metadata", "labels", "kops.k8s.io/instancegroup") == WORKER_NODE_INSTANCEGROUP
+    && node.dig("metadata", "labels", "kops.k8s.io/instancegroup").include?(WORKER_NODE_INSTANCEGROUP)
 end
 
 def get_worker_instance_group_size
   docs = []
   YAML.load_stream(get_kops_config) { |doc| docs << doc }
-  worker_instance_group = docs.last
+  worker_instance_group = docs.last(3)
```
digitalronin, line 89: Per my comment above, can we identify the worker node instance groups via a label?

If we can do that, then we don't need to worry about their position in the YAML file (i.e. `last(3)`). This is less brittle, so we won't break stuff if we reorder the sections of the yaml file.

recycle-node.rb
```diff
@@ -80,19 +80,19 @@ def node_ready?(node)
 
 def worker_node?(node)
   node.dig("metadata", "labels")["kubernetes.io/role"] == "node" \
-    && node.dig("metadata", "labels", "kops.k8s.io/instancegroup") == WORKER_NODE_INSTANCEGROUP
+    && node.dig("metadata", "labels", "kops.k8s.io/instancegroup").include?(WORKER_NODE_INSTANCEGROUP)
 end
 
 def get_worker_instance_group_size
```
digitalronin, line 86: We need to rename this method if it counts workers from multiple node groups. How about `target_worker_node_count` ?

digitalronin (changes requested):

vijay-veeranki:
Hi @digitalronin,

Updated the changes as suggested, can you please look in to it?

poornima-krishnasamy (dismissed):

poornima-krishnasamy (approved):

PR #920, opened 2020-10-08
Increase ingress sleep_delay to 400s

digitalronin:
I dug into the consistently failing test (line 47), and it does pass,
when running locally, with a sufficiently long delay. By repeatedly
testing with different sleep delays, 400s looks like a valid choice to
get tests which pass consistently.

This is a ridiculously long time to wait for a test to run, so I've
added a `speed: "slow"` tag, so that people can bypass this test when
they need faster feedback.


jasonBirchall (approved):
Agreed.

PR #956, opened 2020-10-26
Add a test that we don't get a "Server" header

digitalronin:
Normally, I'd put this test in a spec of its own,
but these tests take so long to set up that it
makes more sense to have multiple expectations in
a single spec.


mogaal (approved):

digitalronin:
depends on https://github.com/ministryofjustice/cloud-platform-infrastructure/pull/955, so I'll merge this after that is merged and applied

PR #985, opened 2020-11-10
Check if the node has ingress controller pod running. This node needs…

poornima-krishnasamy:
… to be checked manually to check connection timeouts before draining


poornima-krishnasamy:
Closing the PR. This is not required.

PR #989, opened 2020-11-16
Add delete node step after draining.

poornima-krishnasamy:
This will allow will remove it from the cluster entirel and as part of that process, it will remove it from ELB targetgroup for de-registering. This will ensure the existing connections are closed from ELB side.

As part of this ticket, the "Connection termination on deregistration" is enabled for the load-balancer of default ingress-controller.

Related to: https://github.com/ministryofjustice/cloud-platform/issues/2540

### DIFF
--- a/recycle-node.rb
+++ b/recycle-node.rb
@@ -42,6 +42,7 @@ def main
   sleep 30
 
   node = get_latest_node_details(node) # node status should have changed, after being drained
+  delete_node(node)
   terminate_node(node)
 
   sleep 60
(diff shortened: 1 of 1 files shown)


jasonBirchall (commented):
If you delete the node (using `kubectl delete nodes`) do you also need to terminate the ec2 instance?

poornima-krishnasamy:
> If you delete the node (using `kubectl delete nodes`) do you also need to terminate the ec2 instance?

@jasonBirchall , yes `kubectl delete node`  will only remove the node from kubernetes side. we need to terminate the ec2 instance so the instance terminates and the scaling will create a new one.

jasonBirchall (approved):

mogaal (approved):

PR #997, opened 2020-11-24
Add argument for adding a Docker config file to nodes

jasonBirchall:
Connects to https://github.com/ministryofjustice/cloud-platform/issues/2566

Due to the rate limiting on un-authenticated user accounts in Docker
hub, we need to add a docker config file to the worker and master nodes.
This is achieved by adding a kops secret, containing the base64 encoded
docker config file.

The following commit amends the `create-cluster.rb` script to allow
passing an argument to include a dockerconfig file or not. If the
argument is passed, the `add_dockerconfig` function checks the file
exists and executes the required command. I didn't want the script to
fail if the dockerfile doesn't exist so it should shield the [add secret](https://github.com/kubernetes/kops/blob/master/docs/cli/kops_create_secret_dockerconfig.md)
command to execute only if it exists.


mogaal (approved):

poornima-krishnasamy (approved):
Looks good

PR #1002, opened 2020-11-30
Updated template kops to include the latest changes

mogaal:


kops/live-1.yaml
```diff
@@ -411,6 +411,11 @@ metadata:
 spec:
   image: kope.io/k8s-1.17-debian-stretch-amd64-hvm-ebs-2020-07-20
   machineType: r5.2xlarge
+  mixedInstancesPolicy:
+    instances:
+    - r5.xlarge
+    - c5.xlarge
+    - r4.xlarge
```
jasonBirchall, line 418: Wrong instances for the "monitoring" nodes

jasonBirchall (changes requested):
Instances need updating.

mogaal:
This was applied in another PR, monitoring instances don't have mixed instances anymore until we fix affinity rules to use label matching instead of instance types matching

PR #1025, opened 2020-12-10
Add "hoodaw_host" variable to EKS components

digitalronin:
This complements "hoodaw_api_key"

Several concourse pipelines post data to HOODAW.
Currently, they get the API key from a secret, but
hard-code the hostname. 

It would be neater if both the API key and the
hostname were available from the same secret.

This change is the first step to doing this. It 
makes the current hostname of the HOODAW web app.
available as a terraform variable.

A later change to the concourse module will put
this value into the kubernetes secret, from where
concourse will be able to retrieve it.


razvan-moj (approved):

razvan-moj (changes requested):
plan fails

razvan-moj (approved):

mogaal (approved):

PR #1038, opened 2020-12-17
Update README.md

mogaal:


terraform/global-resources/README.md
```diff
@@ -8,6 +8,8 @@ These are resources that are global in nature and therefore there are no workspa
 Since resources in multiple accounts are managed here, multiple AWS providers are defined.
 You can see the list of providers in [main.tf](main.tf#L10-L29), as well as the names of the AWS profiles that must be configured for this to run properly.
 
+**You require DSD credentials in order to run a full plan**, you can still apply changes targeting resources. The reason is that in `dns.tf` we are touching DSD resources.
```
digitalronin, line 11: What does this mean in detail? If I have a set of DSD credentials (which I do), how would I need to set things up so that I can do a terraform apply?

digitalronin (changes requested):
Please could you add a bit more detail about how to use the DSD credentials?

terraform/global-resources/README.md
mogaal: You can still do `terraform apply` but only targeting. It means we have some resources in DSD account and when we do full plan terraform will try to check them out. If you don't have the DSD credentials it is going to fail.

digitalronin: Could you put that explanation into the README, please, @mogaal ?

mogaal: @digitalronin I added more information. Do you think is now more clear or we can add something else?

digitalronin: This might just be my lack of knowledge, but I can't see from the README, *how* I would use my DSD credentials, if I had any.

I understand how to supply AWS credentials to terraform, but I don't understand how to supply *multiple sets* of AWS credentials.

If that's just my limited knowledge, and it's obvious to others, then I'm happy to merge this. But if it's an unusual thing to do, I think we should have instructions on how to do it. Something like:

```
You need DSD credentials to apply the terraform code, so to do a full `apply` you would run something like this:

export AWS_ACCESS_KEY_ID=[aws access key for moj-cp]
...
terraform apply -backend-config=...
```

If you don't think that's necessary, then we can merge this as it is now.

jasonBirchall (approved):

jasonBirchall:
After a conversation with the team, it was decided this required more time and thought. Closing this PR and will revisit at a later time.

PR #1062, opened 2021-01-18
fix pf s3 prod

razvan-moj:
as done earlier for preprod


jasonBirchall (approved):

steverendell (changes requested):
There is another diff from preprod - the secret at line 31

PR #1073, opened 2021-01-28
Add a `github_actions_secrets_token` concourse var

digitalronin:
This contains a github personal access token which is able to write to
all MoJ repositories. This is used to create github actions secrets
containing ECR AWS credentials and serviceaccount cluster credentials
via terraform, so that teams don't need to manually copy/paste them in
order to make their CI/CD pipelines work.


digitalronin:
Plan is failing because this version of the concourse module doesn't exist yet.

mogaal (approved):

PR #1083, opened 2021-02-15
Default to encrypt *.tfstate rather than ignoring it

mogaal:
Because we need to find the way how to store the seed state (the one used to create the initial S3/DybamoDB table), it feels a good idea to store it encrypted within the repo instead of ignoring them.


razvan-moj (approved):

razvan-moj (changes requested):
wouldn't that add all the other tfstates to git?

razvan-moj (approved):

PR #1158, opened 2021-05-07
Create live cluster in EKS

jasonBirchall:
Overview
---
This PR connects to https://github.com/ministryofjustice/cloud-platform/issues/2854 and relates to the creation of a new production grade cluster named `live`. This cluster's name is deliberately vague to ensure future upgrades and iterations don't change name. It requires a change to the cluster, main and vpc terraform HCL in this repository. At the end of this PR, we should have a `live` production EKS cluster bootstrapped and initial components installed.

Design choices
---
- I've tried to keep this as close to the current production infrastructure as possible, this includes things like having 19 worker nodes. 
- As we currently have a production EKS cluster called `manager`, we need to manage two production states that require different infrastructure. For example, the `manager` cluster uses only 4 worker nodes (due to its usage), but `live` will require 19. In lieu of a better solution, I've used Terraform conditions to lookup workspace to make these decisions.


terraform/aws-accounts/cloud-platform-aws/vpc/eks/cluster.tf
```diff
@@ -17,6 +17,25 @@ provider "kubernetes" {
   load_config_file       = false
 }
 
+locals {
```
razvan-moj, line 20: this is already in, see main.tf from https://github.com/ministryofjustice/cloud-platform-infrastructure/pull/1201

jasonBirchall:
Merged in different PR's

PR #1179, opened 2021-05-18
Updating Prometheus Operator in Manager 

mogaal:
This should fix the Prometheus operator issue in the manager cluster. It basically reduces the resources quotas set for live-1 (we are in manager, we don't need that amount of memory + we don't have large node groups)

### DIFF
--- a/terraform/aws-accounts/cloud-platform-aws/vpc/eks/components/components.tf
+++ b/terraform/aws-accounts/cloud-platform-aws/vpc/eks/components/components.tf
@@ -106,7 +106,7 @@ module "monitoring" {
   oidc_components_client_secret = data.terraform_remote_state.cluster.outputs.oidc_components_client_secret
   oidc_issuer_url               = data.terraform_remote_state.cluster.outputs.oidc_issuer_url
   enable_thanos_sidecar         = terraform.workspace == local.live_workspace ? true : false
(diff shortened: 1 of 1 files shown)


terraform/aws-accounts/cloud-platform-aws/vpc/eks/components/components.tf
```diff
@@ -106,12 +106,11 @@ module "monitoring" {
   oidc_components_client_secret = data.terraform_remote_state.cluster.outputs.oidc_components_client_secret
   oidc_issuer_url               = data.terraform_remote_state.cluster.outputs.oidc_issuer_url
   enable_thanos_sidecar         = terraform.workspace == local.live_workspace ? true : false
+  enable_large_nodesgroup       = terraform.workspace == local.live_workspace ? true : false
```
poornima-krishnasamy, line 109: @mogaal , this will set `enable_large_nodesgroup=true` for manager as the live_workspace is set to manager in here. 
https://github.com/ministryofjustice/cloud-platform-infrastructure/blob/main/terraform/aws-accounts/cloud-platform-aws/vpc/eks/components/main.tf#L69-L72

mogaal: You are 100% right, fixing it now. Thanks for that.

poornima-krishnasamy (approved):

PR #1195, opened 2021-05-26
Amend eks module to allow for live creation

jasonBirchall:
Overview
---
This PR connects to https://github.com/ministryofjustice/cloud-platform/issues/2854 and relates to the creation of a new production grade cluster named `live`. This cluster's name is deliberately vague to ensure future upgrades and iterations don't change name. It requires a change to the cluster, main and vpc terraform HCL in this repository. At the end of this PR, we should have a `live` production EKS cluster bootstrapped and initial components installed.

What's in this PR
---
We would like to create as close to
`live-1` as possible. `Live-1` is currently running 19 worker nodes and
three masters. At the start of `Live` it will need a different instance
type `r5.xlarge` for mem optimisation but we can keep the instance count
the same and utilise the out of the box autoscaling.

There are also a few tags that need to be included as part of our
tagging strategy.

I wanted to keep the if conditions down as much as possible. This can be
iterated in the future.

### DIFF
--- a/terraform/aws-accounts/cloud-platform-aws/vpc/eks/cluster.tf
+++ b/terraform/aws-accounts/cloud-platform-aws/vpc/eks/cluster.tf
@@ -35,7 +35,7 @@ module "eks" {
       min_capacity     = 1
       subnets          = data.aws_subnet_ids.private.ids
 
-      instance_type = var.worker_node_machine_type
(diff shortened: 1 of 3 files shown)


razvan-moj (approved):

jasonBirchall:
Going to re-create this PR.

PR #1235, opened 2021-06-14
Only enable exporters on live - ONLY merge if throttling is happening

mogaal:
This just in case we face some throttling


jasonBirchall (approved):
Great stuff 👍

mogaal:
This hasn't been a problem for almost two months so I am closing this.

PR #1236, opened 2021-06-15
Dockerfile and makefile for e2e testing

mogaal:
This will be useful when we decide to trigger pipelines that execute the e2e tests. 

- Good news: container can be as small as 40MB
- Bad news: To connect to EKS (not the kops side) you need `aws-cli` installed which adds so many MBs of size

Total image size: 94MB 

https://hub.docker.com/repository/docker/ministryofjustice/cloud-platform-tests/tags?page=1&ordering=last_updated

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,7 +1,6 @@
-FROM golang:1.15-alpine AS builder
+FROM golang:1.16-alpine AS builder
 
-ENV GO111MODULE=on \
-    CGO_ENABLED=0 \
+ENV CGO_ENABLED=0 \
     GOOS=linux
 
 WORKDIR /build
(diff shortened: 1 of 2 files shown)


tests/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+FROM alpine:3.13
+
+RUN apk update && apk add --no-cache --upgrade curl python3 py3-pip
+RUN pip install awscli
+
+RUN apk --purge -v del py-pip
+RUN rm /var/cache/apk/*
+
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.15.1/bin/linux/amd64/kubectl
```
razvan-moj, line 23: this is old, we're at 1.18

tests/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY go.sum .
+RUN go mod download
+
+COPY . .
+RUN cd e2e && go test -c -o e2e-tests 
+
+FROM alpine:3.13
+
+RUN apk update && apk add --no-cache --upgrade curl python3 py3-pip
+RUN pip install awscli
```
razvan-moj, line 18: this is awscli v1, while we use v2 on our laptops and I think all the new niceties will be in v2
There is no Alpine package for it, but https://github.com/aws/aws-cli/issues/4685#issuecomment-829600284 hints at a way to get it built and keep the image small

razvan-moj (changes requested):

tests/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM golang:1.15-alpine AS builder
```
jasonBirchall, line 1: Can we bump this to 1.16?

tests/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ENV GO111MODULE=on \
```
jasonBirchall, line 3: If we agree to go 1.16 we don't need to specify this as it's on by default.
https://blog.golang.org/go116-module-changes

tests/Dockerfile
jasonBirchall, line 23: To keep up to date with latest, can we use:
```
curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
Outlined here: https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/

Unless we need 1.15.1 for a specific reason.

tests/Dockerfile
mogaal: Regarding this, it worked!!!  Thanks for that. The image went down from 96MB to 76MB, but the amount of code added within the Dockefile doesn't feel worth :-(

jasonBirchall (approved):
Nice work 👍

PR #1240, opened 2021-06-17
remove integration ingress

razvan-moj:
```
# rspec ./spec/ingress_controller_spec.rb ./spec/ingress_spec.rb ./spec/modsec_spec.rb --format progress --format documentation

ingress controllers
  default
.    returns 200 for http get
.    does not return a server header
  eks
.    returns 302 for http get
.    does not return a server header

nginx ingress
  when ingress is not deployed
.    fails http get
  when ingress is deployed using 'nginx' ingress controller
.    returns 200 for http get
  when ingress is deployed using 'modsec01' ingress controller
.    responds to http get
  when ingress is deployed with invalid syntax
.    is rejected by the admission webhook

Testing modsec on ingress class: 'modsec01'
  when modsec is deployed
    when the url is benign
.      request succeeds
    when the url is malicious
.      request is blocked
  when modsec is disabled
    when the url is benign
.      request succeeds
    when the url is malicious
.      request succeeds


Finished in 4 minutes 36.8 seconds (files took 0.85491 seconds to load)
12 examples, 0 failures
```

### DIFF
--- a/smoke-tests/spec/ingress_controller_spec.rb
+++ b/smoke-tests/spec/ingress_controller_spec.rb
@@ -1,14 +1,13 @@
 require "spec_helper"
 
-describe "ingress controllers", speed: "fast" do
+describe "test ingress controllers", speed: "fast", "live-1": false do
   context "default" do
-    # This needs to be any url of an ingress on the default controller,
-    let(:url) { "https://reports.cloud-platform.service.justice.gov.uk/dashboard" }
+    host = "prometheus.apps.#{current_cluster}"
+    let(:url) { "https://#{host}" }
 
-    it "returns 200 for http get" do
+    it "returns 302 for http get" do
       result = Net::HTTP.get_response(URI(url))
-      expect(result.code).to eq("200")
-      expect(result.message.strip).to eq("OK")
+      expect(result.code).to eq("302")
     end
 
     it "does not return a server header" do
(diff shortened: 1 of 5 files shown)


smoke-tests/spec/modsec_spec.rb
```diff
@@ -36,7 +36,7 @@
     before do
       annotations_hash = {'nginx.ingress.kubernetes.io/enable-modsecurity': "true"}
       annotate_ingress(namespace, ingress_name, annotations_hash)
-      sleep 30
+      sleep 10
```
jasonBirchall, line 39: Out of curiosity, why the sleep change?

smoke-tests/spec/ingress_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe "ingress controllers", speed: "fast" do
   context "default" do
     # This needs to be any url of an ingress on the default controller,
-    let(:url) { "https://how-out-of-date-are-we.apps.live-1.cloud-platform.service.justice.gov.uk/dashboard" }
+    let(:url) { "https://reports.cloud-platform.service.justice.gov.uk/dashboard" }
```
jasonBirchall, line 6: This won't work on non-live-1 clusters. Ignoring the fact that we will eventually move `reports.cloud.....` onto the new cluster `live`.

jasonBirchall: Perhaps we could use the bootstrap helloworld app?

razvan-moj: Good point, this test is weird - it will not fail, but will not test the test cluster either; it just pings 1 live and 1 manager address

smoke-tests/spec/modsec_spec.rb
jasonBirchall: Just to clarify: you saw no errors on live-1?

jasonBirchall: The reason I ask is because the API usage between live-1 and everything else is VERY different.

razvan-moj: This is all on a test cluster. If live turns out to be slower, happy to create another PR to bump back

smoke-tests/spec/ingress_controller_spec.rb
razvan-moj: helloworld app +interpolation would fail on live because. it doesn't use the 'apps.cluster' hostname

jasonBirchall (approved):

PR #1267, opened 2021-07-01
PSP changes

poornima-krishnasamy:
- Create cluster script : Remove applying psp on fix_psp() as they are moved to terraform and applied before components
-  Add default_allow_privilege_escalation. This is set to false by default and hence the pods which needed privileged psp erred `allowPrivilegeEscalation to false and privileged to true`
- Remove psp yaml files. They are not needed and the psp are converted to terraform resource


vijay-veeranki (approved):

mogaal (approved):

jasonBirchall (commented):
I don't think this will work. How it was previously
after the cluster bootstrapped i.e. installed critical components, it would:
- Remove the default priv psp
- Apply our psps
- Recycle all pods so the bootstrapped pods also picked up the psp

How i've interpreted it to work after this pr
after the cluster bootstrapped i.e. installed critical components, it would:
- Remove the default priv psp
- Recycle all pods so the bootstrapped pods would pick up something we don't want it to (the next psp in the list)
- Install our psps as part of the terraform components layer.

poornima-krishnasamy:
> I don't think this will work. How it was previously
> after the cluster bootstrapped i.e. installed critical components, it would:
> 
> * Remove the default priv psp
> * Apply our psps
> * Recycle all pods so the bootstrapped pods also picked up the psp
> 
> How i've interpreted it to work after this pr
> after the cluster bootstrapped i.e. installed critical components, it would:
> 
> * Remove the default priv psp
> * Recycle all pods so the bootstrapped pods would pick up something we don't want it to (the next psp in the list)
> * Install our psps as part of the terraform components layer.

Thanks @jasonBirchall for looking into this.
Since our PSPs are applied after the cluster bootstrapping, there is no need to reapply at components level. 
So after this PR it will be
- our psps applied along with the cluster bootstrapping using terraform 
- delete eks.privileged psp
- recycle all pods so the bootstrapped pods would pick up the correct psp (ours)
- Install components
Let me know if that would work.

jasonBirchall (approved):

PR #1626, opened 2022-07-13
Remove unused variables from global-resources

jakemulley:
This PR removes all unused variables from this Terraform configuration.

None are actually called (via `var`), hence the removal of `terraform.tfvars` and an empty `variables.tf`.


razvan-moj (dismissed):
something's wrong with tfvars

PR #1656, opened 2022-07-18
Remove unwanted or unused Rpsec tests

jasonBirchall:
Overview
---

This PR removes a number of rspec tests from our suite of tests. The tests removed here have either been made redundant by the move to EKS OR they are covered elsewhere.


smoke-tests/spec/alertmanager_spec.rb
```diff
@@ -1,12 +0,0 @@
-require "spec_helper"
```
poornima-krishnasamy, line 1: Do we have an identical alertmanager check in go?

jasonBirchall: This particular check confirms if the correct servicemonitor exists: We do it slightly differently in Go, checking if services and servicemonitors exist in bulk
https://github.com/ministryofjustice/cloud-platform-infrastructure/blob/main/test/config/config.go#L102-L111

smoke-tests/spec/prometheusrules_spec.rb
```diff
@@ -1,39 +0,0 @@
-require "spec_helper"
```
poornima-krishnasamy, line 1: And this one as well as above? Even if there is a serviceMonitor, should we not check if the rules are present?

smoke-tests/spec/certificates_spec.rb
```diff
@@ -1,12 +0,0 @@
-require "spec_helper"
```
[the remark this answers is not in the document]
jasonBirchall: New test added to cover this

smoke-tests/spec/prometheusrules_spec.rb
jasonBirchall: new test added

poornima-krishnasamy (approved):

PR #1734, opened 2022-07-26
Replace logging Ruby test with Go

jasonBirchall:
Overview
---

This PR replaces one of the last remnants of our Ruby tests. Upon working on this PR, it was clear the test hadn't worked for a long time. The index structure was incorrect, but was reporting a pass.

Here we have a Go replacement utilising the Ginkgo package with Gomega assertions. It saves ~60seconds of runtime.


test/logging_test.go
```diff
@@ -0,0 +1,131 @@
...
+		)
+
+		var (
+			elasticSearch = "https://search-cloud-platform-live-dibidbfud3uww3lpxnhj2jdws4.eu-west-2.es.amazonaws.com"
+			date          = time.Now().Format("2006.01.02")
+			search        = elasticSearch + "/live_kubernetes_cluster" + "-" + date + "/_search"
+		)
+
+		BeforeEach(func() {
+			if !strings.Contains(c.ClusterName, "live") {
```
jakemulley, line 35: This is opinionated, but is it worthwhile using `strings.HasPrefix(c.ClusterName, "live")` here, instead of `strings.Contains`?

I'm thinking about cases where a cluster name may be `non-live`, which would run this test.

jasonBirchall:
I think I've stuffed up the branch here. Will branch off and try again

PR #1852, opened 2022-09-06
allow only adding data in ES logs

razvan-moj:
permit only read actions and adding data, specifically this excludes `es:ESHttpDelete`


jasonBirchall (approved):
I think this looks reasonable.

PR #2174, opened 2023-04-18
feat: 🎸 add vanity url for opensearch cluster

jaskaransarkaria:
tested in ephemeral, see https://et-logs.et.cloud-platform.service.justice.gov.uk/_dashboards

### DIFF
--- a/terraform/aws-accounts/cloud-platform-aws/account/opensearch.tf
+++ b/terraform/aws-accounts/cloud-platform-aws/account/opensearch.tf
@@ -81,6 +81,30 @@ resource "aws_kms_key" "live_modsec_audit" {
   multi_region                       = false
 }
 
+data "aws_route53_zone" "et_cloud_platform_justice_gov_uk" {
+  name = "cloud-platform.service.justice.gov.uk."
+}
+
+# needed for load balancer cert
+module "acm" {
+  source  = "terraform-aws-modules/acm/aws"
+  version = "~> 4.0"
+
+  domain_name = data.aws_route53_zone.et_cloud_platform_justice_gov_uk.name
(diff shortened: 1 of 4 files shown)


terraform/aws-accounts/cloud-platform-aws/account/opensearch.tf
```diff
@@ -149,6 +176,16 @@ resource "aws_opensearch_domain" "live_modsec_audit" {
   }
 }
 
+# add vanity url to cluster 
+resource "aws_route53_record" "opensearch_custom_domain" {
+  zone_id = data.aws_route53_zone.et_cloud_platform_justice_gov_uk.zone_id
+  name    = "et-logs"
```
poornima-krishnasamy, line 182: I think this should be `logs` for the prod account

jaskaransarkaria: good catch

terraform/aws-accounts/cloud-platform-aws/account/opensearch.tf
```diff
@@ -81,6 +81,30 @@ resource "aws_kms_key" "live_modsec_audit" {
   multi_region                       = false
 }
 
+data "aws_route53_zone" "et_cloud_platform_justice_gov_uk" {
```
poornima-krishnasamy, line 84: Just a cosmetic change. Can we name the resource as `cloud_platform_justice_gov_uk`

jaskaransarkaria: good catch

poornima-krishnasamy (approved):

PR #2352, opened 2023-07-13
feat(TF): Add kibana alert for duplicate uid's

jackstockley89:
- feat(TF): kibana alert config for UID duplication
- feat(TF): add provider fpr elasticstack and remove ec resources
- feat(TF): Add tested resources to the global account


jackstockley89:
Moving this into the monitoring module

PR #2455, opened 2023-09-06
Move worker security group outside of eks module

sj-williams:
This PR moves our worker node security group and rules config outside of EKS  module in preparation for 17.24.0 > 18.31.2 EKS module upgrade.

### DIFF
--- a/terraform/aws-accounts/cloud-platform-aws/vpc/eks/security-group.tf
+++ b/terraform/aws-accounts/cloud-platform-aws/vpc/eks/security-group.tf
@@ -41,7 +41,7 @@ resource "aws_security_group_rule" "node_ingress_self" {
 }
 
 resource "aws_security_group_rule" "node_ingress_cluster" {
-  description              = "Allow node pods to receive communication from the cluster control plane."
+  description              = "Allow workers pods to receive communication from the cluster control plane."
   protocol                 = "tcp"
   security_group_id        = aws_security_group.node.id
   source_security_group_id = module.eks.cluster_security_group_id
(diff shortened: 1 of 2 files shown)


terraform/aws-accounts/cloud-platform-aws/vpc/eks/security-group.tf
```diff
@@ -0,0 +1,72 @@
...
+
+resource "aws_security_group_rule" "node_egress_internet" {
+  description       = "Allow nodes all egress to the Internet."
+  protocol          = "-1"
+  security_group_id = aws_security_group.node.id
+  cidr_blocks       = ["0.0.0.0/0"]
+  from_port         = 0
+  to_port           = 0
+  type              = "egress"
+}
```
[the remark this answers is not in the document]
poornima-krishnasamy: This is needed so the nodes allow outbound traffic to all public ips.

terraform/aws-accounts/cloud-platform-aws/vpc/eks/security-group.tf
```diff
@@ -0,0 +1,72 @@
...
+  description              = "Allow workers pods to receive communication from the cluster control plane."
+  protocol                 = "tcp"
+  security_group_id        = aws_security_group.node.id
+  source_security_group_id = module.eks.cluster_security_group_id
+  from_port                = 1025
+  to_port                  = 65535
+  type                     = "ingress"
+}
+
+resource "aws_security_group_rule" "node_ingress_cluster_https" {
```
jaskaransarkaria, line 53: This is covered by the above rule, so is redundant.

sj-williams: we want to keep the same configuration as the existing security group in order to perform the state mv and not recreate any of the resources. The security group needs reviewing as a seperate PR.

jaskaransarkaria (approved):
LGTM

terraform/aws-accounts/cloud-platform-aws/vpc/eks/security-group.tf
poornima-krishnasamy: They are slightly different. The above one is to communicate to each other. and this one is to communicate with the control plane

PR #2457, opened 2023-09-08
perf: ⚡️ bump iops and throughput

jaskaransarkaria:
Your domain is experiencing throttling as you have hit disk throughput limits.


poornima-krishnasamy:
Closing this. This involves a blue-green deployment. Instead of updating tin ES, change will be implemented when moving to Opensearch

PR #2605, opened 2023-11-16
Add tests to check for psa privileged label on system namespaces

poornima-krishnasamy:


poornima-krishnasamy:
Not needed as all labels are set through terraform. If there is any conflict, the divergence would notify

PR #2950, opened 2024-03-27
Bump hashicorp http to 3.4.2

mikebell:


jaskaransarkaria (commented):
The plans are failing because the lock hasn't been updated, push up a tf lock and that should fix the tf errors

jaskaransarkaria (approved):

PR #2995, opened 2024-04-17
Revert PR 2993 for Opensearch Alert

timckt:


timckt:
We need to revert the change because there is an error when applying the change.
```
│ Error: elastic: Error 405 (Method Not Allowed)
│ 
│   with elasticsearch_opensearch_destination.cloud_platform_alerts,
│   on app-opensearch-alerts.tf line 1, in resource "elasticsearch_opensearch_destination" "cloud_platform_alerts":
│    1: resource "elasticsearch_opensearch_destination" "cloud_platform_alerts" {
│ 
╵
```
This maybe related to that destinations will be deprecated going forward in Opensearch from Opensearch Destinations [Page](https://app-logs.cloud-platform.service.justice.gov.uk/_dashboards/app/alerting#/destinations?from=0&search=&size=20&sortDirection=desc&sortField=name&type=ALL).

Also we also need to use [new resource block](https://registry.terraform.io/providers/opensearch-project/opensearch/latest/docs/resources/channel_configuration) `opensearch_channel_configuration` for Opensearch and it involves the testing of new terraform [opensearch provider](https://registry.terraform.io/providers/opensearch-project/opensearch/latest).

PR #3072, opened 2024-05-15
feat: 🎸 add label pods controller to infra and add int tests

jaskaransarkaria:
- move label pods controller to be deployed at infra
- add depends_on to as many components as possible so that the controller can label as many components as possible (the controller will pick up other pods on recreate or update)
- add int tests


poornima-krishnasamy:
@jaskaransarkaria  should we add the label_pods in core layer so it is ready before other components are installed?

poornima-krishnasamy (approved):

PR #3298, opened 2024-08-15
chore: Turn on alert for OpenSearch total node storage

timckt:
We had implemented monitoring the OpenSearch storage size at single node level. This is to turn on alert for monitoring OpenSearch storage size at total node level as well.

According to OpenSearch [CW alarm guide](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/cloudwatch-alarms.html), the alarm should be trigger at 25% of the storage space for each node.
Using this calculation of (size-in-gb * instance count * GB-to-MB * 25% * 1024) for storage threshold. This value is in MiB so need to * 1024.

Also update to use attribute reference instead of hardcoding the storage threshold.

Relates to https://github.com/ministryofjustice/cloud-platform/issues/5928

### DIFF
--- a/terraform/aws-accounts/cloud-platform-aws/account/app-opensearch.tf
+++ b/terraform/aws-accounts/cloud-platform-aws/account/app-opensearch.tf
@@ -391,9 +391,10 @@ module "live_app_logs_opensearch_monitoring" {
   monitor_free_storage_space_total_too_low = true
 
   # Using this calculation of (size-in-gb * 25% * 1024) because 25% is the best-practice for low disk, per AWS's recommendations. This value is in MiB so need to * 1024
-  free_storage_space_threshold = aws_opensearch_domain.live_app_logs.ebs_options[0].volume_size * 0.25 * 1024
-
+  free_storage_space_threshold             = aws_opensearch_domain.live_app_logs.ebs_options[0].volume_size  * 0.25 * 1024
+  
   # Using this calculation of (size-in-gb * instance count * 25% * 1024) because 25% is the best-practice for low disk, per AWS's recommendations. This value is in MiB so need to * 1024
-  free_storage_space_total_threshold = free_storage_space_threshold * aws_opensearch_domain.live_app_logs.cluster_config[0].instance_count
-  tags                               = local.app_logs_tags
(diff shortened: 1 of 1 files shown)


terraform/aws-accounts/cloud-platform-aws/account/app-opensearch.tf
```diff
@@ -383,11 +383,13 @@ resource "auth0_rule_config" "opensearch_app_logs_client_id" {
...
-  sns_topic           = module.baselines.slack_sns_topic
-  min_available_nodes = aws_opensearch_domain.live_app_logs.cluster_config[0].instance_count
-  tags                = local.app_logs_tags
+  source                                   = "github.com/ministryofjustice/cloud-platform-terraform-opensearch-cloudwatch-alarm?ref=0.0.2"
+  alarm_name_prefix                        = "CP-live-app-logs-"
+  domain_name                              = local.live_app_logs_domain
+  sns_topic                                = module.baselines.slack_sns_topic
+  min_available_nodes                      = aws_opensearch_domain.live_app_logs.cluster_config[0].instance_count
+  monitor_free_storage_space_total_too_low = true
+  free_storage_space_total_threshold       = 20480 * aws_elasticsearch_domain.live_1.cluster_config[0].instance_count
```
jaskaransarkaria, line 392: where did the 20480 number come from?

timckt: This is 20GB in MegaByte and we are using this number for ElasticSearch and OpenSearch.

I take a look again, on the [aws guide](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/handling-errors.html#troubleshooting-cluster-block), it mention below

> If one or more nodes in your cluster has storage space less than the minimum value of 
> 1) 20% of available storage space, or 
> 2) 20 GiB of storage space, basic write operations like adding documents and creating indexes can start to fail. 

And according to OpenSearch [CW alarm guide](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/cloudwatch-alarms.html), the alarm should be trigger at 25% or 20GB.

So in our case, as we are using 12000 GB for OpenSerach, and the 25% of it is 3000GB. We better use reference attribute in our case. Will update it.

PR #4163, opened 2025-07-29
Create centralised egress VPC

dms1981:
This PR achieves the following:

* Creates a new VPC - `egress-vpc`
  * This is a simple three-tier VPC with transit, firewall, and public subnets
  * Transit subnets will have TGW endpoints, and routes to the AZ-local firewall endpoint
  * Firewall subnets will have firewall endpoints, and routes relevant to the TGW or AZ-local NAT Gateway
  * Public subnets will have NAT gateways, and routes relevant to the AZ-local firewall endpoint
* Creates a new AWS Network Firewall - `cloud-platform-egress-firewall`
  * This firewall serves as an inline-inspection appliance, checking outbound traffic against AWS Managed Rule Groups in _strict_ order, then allowing all outbound traffic.

A follow-on PR will ensure the VPC is attached to the CP Transit Gateway with appropriate routes.


dms1981:
Putting this into draft until we can test it against Concourse and talk about how to best implement it.

PR #4224, opened 2025-08-19
feat(Go): Build binary to speed up execution time in concourse

jackstockley89:


jackstockley89:
taking a different approach

PR #4255, opened 2025-09-03
Provision Cloud Platform firewall (with routing placeholders)

kyphutruong:
This PR will:

1. Create AWS Network Firewall
2. Create Firewall rule group with allow all traffic
3. Create Firewall policy with the following threat detection:
- [AttackInfrastructureStrictOrder](https://docs.aws.amazon.com/network-firewall/latest/developerguide/aws-managed-rule-groups-atd.html)
- [MalwareDomainsStrictOrder](https://docs.aws.amazon.com/network-firewall/latest/developerguide/aws-managed-rule-groups-domain-list.html)
- [BotNetCommandAndControlDomainsStrictOrder](https://docs.aws.amazon.com/network-firewall/latest/developerguide/aws-managed-rule-groups-domain-list.html)

4. Has commented out placeholders for routing traffic through the firewall

Relates to https://github.com/ministryofjustice/cloud-platform/issues/7451


dms1981 (approved):
LGTM! Only observation would be that you're presently logging both flow and alert logs to the same cloudwatch log group which could get quite chatty, but that can be revisited if requirements for logging / relaying onwards change.

PR #4581, opened 2026-04-10
fix: 🐛 wait for cluster completion before ns label apply

sj-williams:
## Purpose

This PR mitigates the intermittent cluster build failure where dns propogation delays lead to cluster lookup timeout when trying to apply required labels and PSA annotations to system namespaces.

We are adding a `depends_on` to label resources in order to ensure full cluster build completion first. 

relates to ministryofjustice/cloud-platform#8127


tmahmood72 (approved):
LGTM

PR #4613, opened 2026-05-07
chore: 🤖 scale up OLD default node group live-2 and manager

sj-williams:
## Purpose

Step 2 of CP2 node group stabilisation plan

- Scale up OLD `default` ngs in `manager` and `live-2`
- NO-OPS in `live`


tmahmood72 (approved):
LGTM

PR #4647, opened 2026-05-29
Unblock IP Address

dan-criddle:
Re-enabled after 7 days as per previous SOC advice

See [slack thread]( https://mojdt.slack.com/archives/C57UPMZLY/p1779441728743419) for initial report


tmahmood72 (approved):
LGTM

PR #4709, opened 2026-08-27
test: add test for kube-state-metrics prometheus job status

JakeBroughton:
Adds an integration test to validate if kube-state-metrics is healthy and scraping metrics.

Adding prior to the kube-state-metrics 2.18 update https://github.com/ministryofjustice/cloud-platform/issues/8357


dan-criddle (dismissed):
Possible improvement: could the test try again if the metrics series isn't found (line 50)? May improve the chances of it passing if Prometheus is slow. Not a blocker though, approving.

dan-criddle (approved):

PR #4717, opened 2026-09-16
feat: widen Container Platform route to the nonlive range

FolarinOyenuga:
feat: Replaces the single octo-nonlive VPC (10.195.48.0/20) with the full Container Platform nonlive range, so the other nonlive spokes do not each need a separate entry later.

Pairs with https://github.com/ministryofjustice/deployment-tgw/pull/340, which routes the same range back from the MoJ transit gateway to the Modernisation Platform peering. Both sides need to match or live-1 will not send to the other spokes.

Container Platform live spokes are in 10.41.0.0/16 and are not included, just nonlive for now.

Ref: https://github.com/ministryofjustice/cloud-platform/issues/8371


terraform/aws-accounts/cloud-platform-aws/vpc/transit-gateway-routes/routes.tf
```diff
@@ -20,7 +20,7 @@ locals {
     "10.162.96.0/20",  # Delius Core Training VPC
     "10.26.0.0/16",    # Mod Platform Dev and Test
     "10.27.0.0/16",    # Mod Platform Pre-Prod and Prod
-    "10.195.48.0/20",  # Container Platform octo-nonlive
+    "10.195.0.0/16",   # Container Platform nonlive
```
davidkelliott:
```suggestion
    "10.195.0.0/16",   # Container Platform

tmahmood72 (dismissed):
LGTM

davidkelliott (changes requested):

davidkelliott (approved):

